PR #1, opened 2019-03-19
Changed the target trajectory from 2D to 3D

KamranBinaee:
The first version of my fork and I only made a minor change in the way the target position is being recorded (2D -> 3D)


jspjutNV:
Looking at this, I'm confused why we'd want to move from the azimuth and elevation parameterization that Joohwan had asked for to the cartesian x,y,z in this pull request. Was there a change in direction? @joohwankimNV

jspjutNV:
Looking closer, maybe this is just a change to how we log the data? Is there a reason we shouldn't log both formats to avoid having to recompute the transformation later?

KamranBinaee:
So the reason is, moving from angular space to 3D target position would be
impossible since we lose depth value. It’s always easy to calculate the az,
el during analysis and only record the raw values in DB files. If in the
future we wanted to reconstruct the data we’d have all we need.
Let me know what you think?

KamranBinaee:
This is just simply the target position wrt head (view point) therefore
calculating the radius of rotation for the target should be super easy.
Similarly transforming to spherical (angular) coordinate system.
For aim though, all we have is a normalized direction vector, therefore
it’s fine to keep it as (az,el)

joohwankimNV:
Sorry I did not know how to reply in the correct way. Copying my email message here:

Sorry for the confusion - I did not update with the rest of the group regarding a conversation I had with Kamran. I agreed to going to 3D cartesian points as that is the complete information for spatial relationship. Josef, you are right that we could have solved the same problem by recording target distance. But let's settle down on what Kamran did for now.

Kamran, could you also reflect the changes in the schema spreadsheet so that all information is on the same page?

KamranBinaee:
Sure,
I just updated the progress sheet.
Thanks

code/source/App.cpp
```diff
@@ -372,14 +372,11 @@ Point2 App::getViewDirection()
 	return Point2(az, el);
 }
 
-Point2 App::getTargetDirection()
```
jspjutNV: We could keep `App::getTargetDirection()` in it's old form and just add `App::getTargetPosition()` as implemented here. This gives us more flexibility to switch which function we use later.

code/data-files/experimentconfig.Any
```diff
@@ -2,8 +2,8 @@
     "settingsVersion" : 1,
     "taskType" : "target",//"target","reaction",
     "expMode" : "real",
-    "expVersion" : "ComplexMotion",
```
jspjutNV: I don't think we need the changes in this file `experimentconfig.Any` or `userconfig.Any`.

code/maxPerf.vcxproj
```diff
@@ -22,7 +22,7 @@
     <VCProjectVersion>15.0</VCProjectVersion>
     <ProjectGuid>{D0B15FD1-8D51-4033-B19F-477FAAF59787}</ProjectGuid>
     <RootNamespace>maxPerf</RootNamespace>
-    <WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>
```
jspjutNV: We could switch `WindowsTargetPlatformVersion`, but we might want to check with everyone before pushing this change to the `master` branch.

jspjutNV:
FYI, this is a generally good changelist. Joohwan and I are just using this as an opportunity to learn this system. I do think it would be good to remove the changes to the `.Any` files from the pull request.

morgan3d:
Changing the windows platform version creates all kinds of obscure linker
problems if everyone doesn't have exactly the same Windows version
installed...and it can create problems with the statically linked
libraries. So, I recommend having two or three people test that one
immediately after push and be prepared to roll back.

-m

code/source/App.cpp
joohwankimNV: I agree with Josef's suggestion here. A reminder that this also requires changes in where the logging is done.

joohwankimNV:
Hey Kamran, could you process the changes requested by Josef?

KamranBinaee:
> Hey Kamran, could you process the changes requested by Josef?

Yes, Since this process is also new to me. I'm figuring out the easiest way to apply the changes. Or maybe it'd be easier to go for another pull request? Not sure.

joohwankimNV:
Let's keep pushing with this pull-request. It is a learning experience for all of us. I notice you made a change in experimentConfig.Any. In my understanding there are three more things to do:
- Exclude change in userConfig.Any (for the same reason as with experimentConfig.Any)
- Exclude change in windows version. Let's do this as a new pull-request after completing this one. That way we can only revert the last version if window version creates any problem.
- Keep both methods of position logging: angular position and cartesian position. Note that we need to reflect this change in other parts of the code as well.

Issue #23, opened 2019-06-27
Target Health Bar


bboudaoud-nv:
There is commented our source for a target health bar (drawn in a fixed position in screen space) at [line 847 of App.cpp](https://github.com/NVlabs/abstract-fps/blob/master/target/source/App.cpp#L857) do we want to leave this here as a reference or should we remove it since the target changes color? Do we want an option in the [experiment config file](https://github.com/NVlabs/abstract-fps/blob/master/target/data-files/experimentConfigReadme.md) to control this?

jspjutNV:
I like keeping around the health bar and creating an issue to make it track the target as you'd expect.

Adding a config option to allow this kind of health indicator to be enabled or disabled sounds appropriate as well. Default value should be `False`.

While we're at it, we might want to add a config option to enable target health color shift which can be set to `True` by default.

bboudaoud-nv:
I (broadly) agree with this, and could see us offering up 3 options here:

1. The current colored target (maybe w/ customizable colors eventually)
2. A health bar that tracks the target
3. A health bar that is positioned statically somewhere in screen space

I think having 2 flags could manage this. We could also use the lack of a provided position for the health bar as an indicator to tie it to the target. As for color shift, I was debating whether we let the user just specify 2 colors (max/min health) and then let the app interpolate between the two. If the two colors are equal then the target is just constant color.

jspjutNV:
I don't think a static target health bar makes sense very often. I can think of a few "boss fight" type scenarios in games where they add a health bar in the center top of the screen. Here's an example screenshot of that.
![image](https://user-images.githubusercontent.com/26633010/60290576-07f6e880-98e7-11e9-80c6-da6559ddd38f.png)

Key to these kinds of scenarios is that the boss often has many targets that are valid and all contribute toward damage to the same health pool. While implementing this style of health bar is easier, I don't know that we need it to be a core feature we maintain like the target color or tracked health bar.

bboudaoud-nv:
I'm not sure we don't want to support the option (especially since it is trivial once you have the option for positioning it above the target).

Since @jspjutNV posted the image above, do we also want to start another issue (possibly referencing this one) for **floating combat text**? I know we discussed this at some point, but weren't sure how we wanted to implement it.

bboudaoud-nv:
[...]

PR #33, opened 2019-07-13
Draws health bars above targets

pknowlesnv:
Adds TargetEntity base class with health, replacing App::m_targetHealth.

Renders health bars above targets and adds a toggle for it to the config.

NVlabs/abstract-fps#23


bboudaoud-nv:
I've gone ahead and manually merged this work in as of [a recent commit](https://github.com/NVlabs/abstract-fps/commit/4eb737dbd20525a18da6c8e5938f256cf7de420a). You can review as you see fit, but mostly I've kept your methods the same. If there's anything important I've missed just let me know.

PR #101, opened 2020-02-13
Added first pass hit decal ("spark")

jspjutNV:


jspjutNV:
@joohwankimNV you should pull this branch and test it out to see if you like the appearance.

source/App.cpp
```diff
@@ -1543,28 +1585,31 @@ void App::onUserInput(UserInput* ui) {
...
 							}
-							WeaponConfig& wConfig = sessConfig->weapon;
 							// Draw a decal here if we are in hitscan mode
-							if (wConfig.hitScan && wConfig.renderDecals && hitDist < finf()) {
-								CFrame frame = activeCamera()->frame();
+							if (wConfig.hitScan && hitDist < finf()) {
 								// Draw decal at the lookRay/world intersection
-								Point3 position = frame.translation + frame.lookRay().direction() * (hitDist - 0.01f);
-								drawDecal(position, info.normal);
+								drawDecal(info.point + 0.01f*info.normal, info.normal);
```
jspjutNV: The old code was a hack without comment. Might want to add a comment here.

source/App.h
```diff
@@ -98,8 +98,11 @@ class App : public GApp {
 	Array<shared_ptr<FloatingCombatText>>	m_combatTextList;					///< Array of existing combat text
 
 	shared_ptr<Weapon>						m_weapon;							///< Current weapon
-	shared_ptr<ArticulatedModel>			m_decalModel;						///< Model for the miss decal
-	Array<shared_ptr<VisibleEntity>>		m_currentDecals;					///< Storage for miss decals
+	shared_ptr<ArticulatedModel>			m_missDecalModel;					///< Model for the miss decal
+	shared_ptr<ArticulatedModel>			m_hitDecalModel;					///< Model for the hit decal
+	shared_ptr<VisibleEntity>				m_hitDecal;							///< Pointer to hit decal
```
jspjutNV: Spacing looks weird here. Check if tabs are doing something wrong...

jspjutNV:
Other than the comments I left, I think this change looks good. Just waiting for Joohwan to approve the look.

source/App.h
bboudaoud-nv: Hmm, odd this looks correct in VS. It might also be time to make a tabs vs spaces determination for FPSci.

source/App.cpp
bboudaoud-nv: This is done 4 places throughout `App.cpp`. Does it make sense to just add a comment to each location this is done? 

Does something along the lines of:
```
// Offset slightly along normal to avoid Z-fighting the wall
```
make sense?

jspjutNV: In case the "+1" wasn't clear enough, I support the use of this comment throughout.

source/App.h
jspjutNV: Given that G3D uses "spaces as tabs" we probably want to use spaces everywhere. Not sure how most repositories add style guides, but we should include something with the repo.

bboudaoud-nv: Okay, I think this belongs in a separate commit/pull request. I'd suggest we leave this as-is and come back around to it in another item. 

Do we want to create a github issue to track converting to spaces? Should we do anything to enforce this on check-in (other than use of pull requests)?

jspjutNV: Yeah, the full convert to spaces should be a separate issue/pull request. I'm okay with adding an "action"

source/App.cpp
bboudaoud-nv: @jspjutNV comments have been added. Feel free to resolve this if they are sufficient.

joohwankimNV:
Looking good! Thanks you!

PR #102, opened 2020-02-21
Improve target spawning

bboudaoud-nv:
This is a branch to simplify the interface(s) used to spawn targets from the app/session. Currently this branch is a work in progress.


bboudaoud-nv:
This is in a reasonable place for merge. We can always continue this refactor in another branch if we didn't "get everything".

PR #103, opened 2020-03-05
Mouse sensitivity change when scoped

bboudaoud-nv:
These changes add support for changing mouse sensitivity when in scoped mode. There are 2 mechanisms to do this:

1.  By default the mouse sensitivity is scaled by the ratio of the (horizontal) scope field of view and the "normal" horizontal field of view. This results in a lower sensitivity when the scope field of view is reduced beyond the "normal" field of view.

2. The use can override the behavior above by specifying the `scopeTurnScale` in the `userconfig.Any` file. When this `Vector2` is not 0 length (i.e. not equal to `Vector2(0,0)`) the turn scale here applies (directly) when in scoped mode. Note that this compounds with the session config and player config `turnScale`s.

### DIFF
--- a/docs/userConfigReadme.md
+++ b/docs/userConfigReadme.md
@@ -11,14 +11,17 @@ The [`userconfig.Any` file](../data-files/userconfig.Any) is located in the [`da
 # User Table
 Each entry in the user table contains the following fields:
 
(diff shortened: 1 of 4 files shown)


joohwankim:
Great! Works as desired.

PR #105, opened 2020-03-20
Improvements for bullet/explosion parametric control

bboudaoud-nv:
This branch includes development of additional controls for rendered "bullets" and explosions (aka target destroy decals). This comes from a request from @joohwankimNV suggesting issues that are arising from new use cases.

Things to Do Include
---
* Bullets
  * Size control
  * Start point/offset control
  * _Color control_
* Explosions
  * Tie explosion time to simulation time (`sdt` approach)
  * Allow user customization of "destroy decal" (size/texture/display time) _on a per-target basis_

### DIFF
--- a/source/App.cpp
+++ b/source/App.cpp
@@ -168,12 +168,16 @@ void App::loadModels() {
 	}
 
 	// Add all the unqiue targets to this list
-	Table<String, Any> toBuild;
+	Table<String, Any> targetsToBuild;
+	Table<String, String> explosionsToBuild;
+	Table<String, float> explosionScales;
 	for (TargetConfig target : experimentConfig.targets) {
-		toBuild.set(target.id, target.modelSpec);
+		targetsToBuild.set(target.id, target.modelSpec);
+		explosionsToBuild.set(target.id, target.destroyDecal);
+		explosionScales.set(target.id, target.destroyDecalScale);
 	}
 	// Append the basic model automatically (used for reference targets for now)
-	toBuild.set("reference", PARSE_ANY(ArticulatedModel::Specification{
+	targetsToBuild.set("reference", PARSE_ANY(ArticulatedModel::Specification{
 		filename = "model/target/target.obj";
 		cleanGeometrySettings = ArticulatedModel::CleanGeometrySettings{
 					allowVertexMerging = true;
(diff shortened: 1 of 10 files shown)


bboudaoud-nv:
I wasn't able to find anything specific to the issue of bullet offset from the wall in non-hitscan mode the way Joohwan described. We should discuss this more, but I'm not sure it should hold this pull request up.

joohwankim:
All the requested features are working well!

bboudaoud-nv:
It seems like we might develop the additional feature of allowing per-target hit decals to be specified in this branch as well.

jspjutNV:
@bboudaoud-nv , is this pull request now a work in progress to add in the per-target hit decal work?

source/App.cpp
```diff
@@ -168,12 +168,16 @@ void App::loadModels() {
 	}
 
 	// Add all the unqiue targets to this list
-	Table<String, Any> toBuild;
+	Table<String, Any> targetsToBuild;
+	Table<String, String> explosionsToBuild;
+	Table<String, float> explosionScales;
 	for (TargetConfig target : experimentConfig.targets) {
```
jspjutNV: I'm noticing that in this loop, if multiple targets have the same `id` then they'll overwrite each other in the table. Does the Any reader report duplicate target `id`s as an error?

source/GuiElements.cpp
```diff
@@ -318,22 +318,22 @@ WeaponControls::WeaponControls(WeaponConfig& config, const shared_ptr<GuiTheme>&
 		n->setWidth(300.0f);
 		n->setUnitsSize(50.0f);
 	} pane->endRow();
-	pane->beginRow(); {
```
jspjutNV: Is removing the Muzzle offset GUI related to this pull request?

source/TargetEntity.h
```diff
@@ -156,6 +156,10 @@ class TargetEntity : public VisibleEntity {
 		return pow(1.0f + TARGET_MODEL_ARRAY_SCALING, m_scaleIdx - TARGET_MODEL_ARRAY_OFFSET);
 	}
 
+	String id() {
```
jspjutNV: should this be `const String &id() {`?

source/App.cpp
bboudaoud-nv: Targets are/have historically been assumed to be unique on `id` (this is the only field by which they are referenced in the `trials` array within the session list). 

Currently we do not throw an exception if a target ID is duplicated, but rather just use the first instance in the array. We can certainly add an exception (at experiment config load time) for repeat target entries.

source/GuiElements.cpp
bboudaoud-nv: Commenting out muzzle offset (from all code) was agreed on as one of the todos for this branch. These fields exist(ed), but weren't connected to anything inside of the application so they were removed.

source/TargetEntity.h
bboudaoud-nv: Sure, moreover it should _really_ be:  `const String& id() const {`

source/App.cpp
bboudaoud-nv: I've created [an issue](https://github.com/NVlabs/abstract-fps/issues/109) to track this. The solution will likely be developed in a new branch.

source/TargetEntity.h
bboudaoud-nv: [This commit](https://github.com/NVlabs/abstract-fps/pull/105/commits/6c3190cc1283e8cdaf0dee19ce9dc7b473604a6d) improves the use of `const` around `TargetEntity` member methods.

bboudaoud-nv:
@jspjutNV I suggest we leave any target-side improvements for hit decal for another branch since this branch was primarily focused on bullet and explosion parametric control. If you agree we can likely close this pull request.

jspjutNV:
Since @joohwankimNV already said this looks good, I'm going to merge it. Nice work.

PR #108, opened 2020-03-26
User settings menu window improvements

bboudaoud-nv:
This is a place to develop improvements to the user menu displayed by the application when Tab is pressed (this applies regardless of the state of `developerMode` in `startupconfig.Any`).


bboudaoud-nv:
Opening this up for review to get more opinions from others. Probably still have some work to do before merge.

I notice that the menu still seems to have a small red circle/"X" in the top left, seems like this shouldn't be there for a `MENU_WINDOW_STYLE`?

source/App.cpp
```diff
@@ -100,14 +99,7 @@ void App::updateMouseSensitivity() {
     // rad/dot = rad/cm * cm/dot = 2PI / (cm/turn) * 2.54 / (dots/in) = (2.54 * 2PI)/ (DPI * cm/360)
     const double mouseSensitivity = 2.0 * pi() * 2.54 / (user->cmp360 * user->mouseDPI);
     const shared_ptr<FirstPersonManipulator>& fpm = dynamic_pointer_cast<FirstPersonManipulator>(cameraManipulator());
-    if (m_userSettingsMode) {
```
jspjutNV, line 103: Removing this, combined with switching to calling `App:setDirectMode(...)` from the `UserMenu` has changes the control of the mouse input mode from the app to being somewhat implicitly tied to the `UserMenu` state. One problem I had when trying to merge the latest master into this branch was that VS refused to "find all references" to `App::setDirectMode()` for me, so I thought it was an unused function and tried to remove it. Only after that failed to compile did it realize where it was being called from inside the `UserMenu`.  I know this old way of changing modes wasn't exactly the best, but maybe it's worth bringing back this block and instead checking the `UserMenu::visible()` state instead of the (now removed) `m_userSettingsMode` that was redundant with the state of the menu visibility. This way, we could (if we wanted to) add in other menu states to part of the decision making around what type of input to use.

jspjutNV:
I generally like the state of this pull request right now, but since the mode when `allowReticleChange = true;` only has a cosmetic bug, I'll open an issue for that problem and just merge this change now.

Issue #109, opened 2020-03-27
Multiple targets with the same ID in experimentconfig.Any


bboudaoud-nv:
Currently there is no protection against an experiment developer specifying multiple targets with the same `id` field value. 

As an additional layer of this problem, the _first_ target specified controls the target parameter behavior (w/ the exception of the models/materials). The _last_ target specified ends up controlling the target model since the dictionary approach used for model storage relies on the target `id` being unique.

We should consider developing a system to check that every element in the `targets` array has a unique `id` at load time.

bboudaoud-nv:
Pull Request #110 addresses this issue with a quick fix for throwing an exception at `experimentconfig.Any` load time.

bboudaoud-nv:
#110 has been merged. This issue is closed.

PR #110, opened 2020-03-27
Don't allow the same Target ID to be entered multiple times

bboudaoud-nv:
Currently an experiment designer can specify two different targets with the same `id` field value in the `experimentconfig.Any` file.

This was described in Issue #109. This branch fixes this problem by raising an exception at `experimentconfig.Any` load time if any two specified targets share a common `id`.


jspjutNV:
Looks great.

PR #114, opened 2020-04-03
Weapon Refactor

bboudaoud-nv:
Refactor the `Weapon` class code to better split it's functionalities from `App`/have it manage it's own projectiles/decals.

### DIFF
--- a/source/App.cpp
+++ b/source/App.cpp
@@ -67,8 +67,7 @@ void App::onInit() {
 	scene()->registerEntitySubclass("FlyingEntity", &FlyingEntity::create);			// Create a target
 
 	m_weapon = Weapon::create(std::make_shared<WeaponConfig>(experimentConfig.weapon), scene(), activeCamera());
-	using namespace std::placeholders;
-	m_weapon->setHitCallback(std::bind(&App::hitTarget, this, _1));
+	m_weapon->setHitCallback(std::bind(&App::hitTarget, this, std::placeholders::_1));
 	m_weapon->setMissCallback(std::bind(&App::missEvent, this));
 
 	// Load models and set the reticle
--- a/source/App.cpp
+++ b/source/App.cpp
@@ -1238,17 +1238,11 @@ void App::onUserInput(UserInput* ui) {
 	for (GKey dummyShoot : keyMap.map["dummyShoot"]) {
 		if (ui->keyPressed(dummyShoot) && (sess->presentationState == PresentationState::feedback)) {
 			Array<shared_ptr<Entity>> dontHit;
-			//dontHit.append(m_currentMissDecals);
 			dontHit.append(m_explosions);
-			//for (auto projectile : m_projectileArray) { dontHit.append(projectile.entity); }
 			Model::HitInfo info;
 			float hitDist = finf();
 			int hitIdx = -1;
 			shared_ptr<TargetEntity> target = m_weapon->fire(sess->targetArray(), hitIdx, hitDist, info, dontHit);			// Fire the weapon
-			//if (notNull(target)) {
-			//	// If we hit a target, destroy it
-			//	sess->destroyTarget(hitIdx);
-			//}
 		}
 	}
 
--- a/source/App.cpp
+++ b/source/App.cpp
@@ -82,10 +82,6 @@ void App::onInit() {
 	outputFont = GFont::fromFile(System::findDataFile("arial.fnt"));
 	hudTexture = Texture::fromFile(System::findDataFile("gui/hud.png"));
 
-	// Load models and set the reticle
-	loadModels();
-	setReticle(userTable.getCurrentUser()->reticleIndex);
-
 	updateMouseSensitivity();			// Update (apply) mouse sensitivity
 	updateSessionDropDown();			// Update the session drop down to remove already completed sessions
 	updateSessionPress();				// Update session to create results file/start collection
--- a/source/App.cpp
+++ b/source/App.cpp
@@ -66,7 +66,14 @@ void App::onInit() {
 	scene()->registerEntitySubclass("PlayerEntity", &PlayerEntity::create);			// Register the player entity for creation
 	scene()->registerEntitySubclass("FlyingEntity", &FlyingEntity::create);			// Create a target
 
-	m_weapon = Weapon::create(std::make_shared<WeaponConfig>(experimentConfig.weapon), scene(), activeCamera(), &m_projectileArray);
+	m_weapon = Weapon::create(std::make_shared<WeaponConfig>(experimentConfig.weapon), scene(), activeCamera());
+	using namespace std::placeholders;
+	m_weapon->setHitCallback(std::bind(&App::hitTarget, this, _1));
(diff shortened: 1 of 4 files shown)


source/App.cpp
```diff
@@ -1327,32 +1195,21 @@ void App::onUserInput(UserInput* ui) {
 					if (sess->canFire()) {
 						fired = true;
 						sess->countClick();														// Count clicks
-						Array<shared_ptr<Entity>> dontHit = { m_hitDecal };
-						dontHit.append(m_currentMissDecals);
+						Array<shared_ptr<Entity>> dontHit;
 						dontHit.append(m_explosions);
-						for (auto projectile : m_projectileArray) { dontHit.append(projectile.entity); }
```
jspjutNV, line 1333: I notice the `dontHit` array no longer includes projectiles, hit decals and miss decals. Is there a reason for this?

source/App.cpp
```diff
@@ -1381,26 +1238,17 @@ void App::onUserInput(UserInput* ui) {
 	for (GKey dummyShoot : keyMap.map["dummyShoot"]) {
 		if (ui->keyPressed(dummyShoot) && (sess->presentationState == PresentationState::feedback)) {
 			Array<shared_ptr<Entity>> dontHit;
-			dontHit.append(m_currentMissDecals);
+			//dontHit.append(m_currentMissDecals);
 			dontHit.append(m_explosions);
-			for (auto projectile : m_projectileArray) { dontHit.append(projectile.entity); }
+			//for (auto projectile : m_projectileArray) { dontHit.append(projectile.entity); }
```
jspjutNV, line 1243: again we removed a bunch of stuff from the `dontHit` list, seems like this might cause collisions with miss decals and projectiles that we didn't want.

source/Weapon.cpp
```diff
@@ -17,21 +96,121 @@ void Weapon::onPose(Array<shared_ptr<Surface> >& surface) {
...
+				m_hitCallback(closestTarget);
+				// Offset position slightly along normal to avoid Z-fighting the target
+				drawDecal(info.point + 0.01 * info.normal, m_camera->frame().lookVector(), true);
+				projectile->clearRemainingTime();
+			}
+			// Handle (miss) decals here
+			else {
+				Array<shared_ptr<Entity>> dontHitItems = dontHit;
+				dontHitItems.append(m_currentMissDecals);
+				dontHitItems.append(targets);
```
jspjutNV, line 135: avoiding hitting targets seems a little weird here since a miss should have missed the targets in the first place. It probably isn't a problem to skip them, but they also shouldn't be hit here either since it's supposed to be a miss case.

source/Weapon.cpp
```diff
@@ -17,21 +96,121 @@ void Weapon::onPose(Array<shared_ptr<Surface> >& surface) {
...
+				const Ray ray = projectile->getDecalRay();
+				m_scene->intersect(ray, closest, false, dontHitItems, info);
+
+				// If we are within 2 simulation cycles of a wall, create the decal
+				if (closest < hitThreshold) {
+					// Offset position slightly along normal to avoid Z-fighting the wall
+					drawDecal(info.point + 0.01 * info.normal, info.normal);
+					projectile->clearRemainingTime();							// Stop the projectile here
+					m_missCallback();
+					//sess->accumulatePlayerAction(PlayerActionType::Miss);		// Declare this shot a miss here
```
jspjutNV, line 148: Should this commented out line stay here?

source/Weapon.cpp
jspjutNV: This is probably correct, I just don't understand the context of what is being hit or missed here and why we're trying to avoid some things in the intersect call. Maybe this just means we need better comments describing these situations in more detail?

source/Weapon.h
```diff
@@ -2,69 +2,105 @@
...
+	Projectile(const shared_ptr<VisibleEntity>& e, float velocity, bool collision = true, float gravity = 0.0f, SimTime t = 5.0) :
+		VisibleEntity(*e),
 		m_totalTime(t),
 		m_collision(collision),
 		m_velocity(velocity),
-		m_gravity(gravity){}
+		m_gravity(gravity) {}
+
+public:
+	Projectile() : m_totalTime(0) {}
```
jspjutNV, line 25: does this constructor need to be public, or could it be protected?

source/Weapon.h
```diff
@@ -131,25 +135,13 @@ class Weapon : Entity {
...
-
-protected:
-	Weapon(shared_ptr<WeaponConfig> config, shared_ptr<Scene>& scene, shared_ptr<Camera>& cam, Array<Projectile>* projectiles) : 
-		m_config(config), m_scene(scene), m_camera(cam), m_projectiles(projectiles) {};
-
-	shared_ptr<ArticulatedModel>    m_viewModel;						///< Model for the weapon
-	shared_ptr<ArticulatedModel>    m_bulletModel;						///< Model for the "bullet"
-	shared_ptr<Sound>               m_fireSound;						///< Sound for weapon firing
-	shared_ptr<AudioChannel>		m_fireAudio;						///< Audio channel for fire sound
+	//void setProjectiles(Array<Projectile>* projectileArray) { m_projectiles = projectileArray; };
```
jspjutNV, line 138: Should this commented out function stay?

source/App.cpp
bboudaoud-nv: This now gets handle at [line 133 of `Weapon::simulateProjectiles()`](https://github.com/NVlabs/abstract-fps/pull/114/files#diff-bd4b3d1e04cccc48e9bbdeb92de4b6c8R133) we only need to add the things that the weapon doesn't own/isn't passed here.

We pass in the target array, so for now it's just explosions that remain belonging to the `App`.

source/App.cpp
bboudaoud-nv: See the [comment above](https://github.com/NVlabs/abstract-fps/pull/114#discussion_r403176424) for why we don't add other things here.

We should verify that there aren't issues with miss decals/multiple projectiles in flight simultaneously though.

source/Weapon.cpp
bboudaoud-nv: I don't think we "need" to add the targets to the don't hit list here (we are just getting the normal for the miss decal placement) we assume targets aren't something we need to worry about at this stage.

Comments could definitely be improved.

source/Weapon.cpp
bboudaoud-nv: No, more of a reminder to thoroughly test player action logs. Can be removed.

source/Weapon.h
bboudaoud-nv: I believe I got a compiler complaint when this wasn't public. 

I don't know the exact error off-hand, but I believe it had to do with construction. Though this might just be as simple as missing a void `static shared_ptr<Projectile> create()` for Array to initialize an `Array<shared_ptr<Projectile>>` from.

source/Weapon.h
bboudaoud-nv: No it should be removed, will remove in a clean-up pass.

PR #118, opened 2020-04-20
Split Player Camera from Debug Camera and Restore G3D Scene Editor Functionality

### LINKED ISSUE #116: The DebugCamera should be independently controllable
G3D apps like to have a separate `DebugCamera` that you can switch to for debugging purposes. We should change the way FPSci handles the player camera to decouple the `DebugCamera` from the player's view and camera.

bboudaoud-nv:
We have been using `activeCamera()` for our player camera, which when defaulting to a debug camera removes the ability to use the developer window/F2 to move to a free-moving camera in the scene. Here we add in a `playerCamera` intended to split this functionality again.

All player-based actions (aside from waypoint editing) now use the player camera, leaving the debug/additional cameras free for future implementation.

Merging this closes #116.


bboudaoud-nv:
Once complete, this should resolve issue #116. Still some work to do in getting a `FirstPersonManipulator` to control the debug camera.

bboudaoud-nv:
Upon discussion w/ @morgan3d I think we may not be using the `FirstPersonManipulator` as intended. It seems we should consider making some changes here/that G3D may make some changes to make it more explicit the default `FirstPersonManipulator` provided by `GApp` should only be used with the debug camera.

This (likely) means we should be constructing our own manipulator to use in order to switch between mouse-as-cursor and mouse-as-camera control mode.

jspjutNV:
Oh, okay, so we should do a `FirstPersonManipulator::create()` call somewhere and keep that along side the default one provided by `GApp`? I assume there should also be changes made to the input handling to change the routing of input depending on which camera/manipulator is active.

jspjutNV:
This latest merge broke the testing. We should make sure there's no regression on testing before we consider this pull request ready.

bboudaoud-nv:
I got the FPM moving the debug camera (we didn't copy over this line from the `GApp::onSimulation()` call into our `FPSciApp::onSimulation()` (which doesn't call `GApp::onSimulation()`, but instead reproduces a subset of the calls from within it).

This PR can now be reviewed.

jspjutNV:
The same 4 tests that are passing in master are passing again with the latest commit.

jspjutNV:
I believe this is ready to merge with the exception of the crash that is now documented in #177 . This bug probably isn't added by this pull request and instead is likely a bug that was always present but wasn't visible because the larger bug this pull request solves was hiding it.

Issue #113, opened 2020-04-02
`App::onSimulation` calls `scene()->onSimulation` twice


jspjutNV:
In `App::onSimulation` in `App.cpp` this line happens twice in a row, but should only happen once.

```if (scene()) { scene()->onSimulation(sdt); }```

While the code fix for this is straightforward, I want to open this issue first to allow discussion to make sure everyone is ready for the fix once we deploy it.

jspjutNV:
The effect of this bug is that everything being simulated by the scene is effectively running twice as fast as intended. I tested removing the duplicate line and it seemed like all target motion was half as fast, which would mean experiments would need to be redesigned.

bboudaoud-nv:
This issue was closed by the merge of #119.

PR #119, opened 2020-04-21
Remove extra call to scene onSimulation()

bboudaoud-nv:
Addresses issue #113


jspjutNV:
This is a good change to make. We're going to double all the default config parameters that depended on the sdt since they were being run twice before, effectively doubling things like speed.

PR #121, opened 2020-04-22
Mouse sensitivity update

### LINKED ISSUE #120: Mouse Sensitivity Issue from PlayerEntity->onSimulation()
The use of angular velocity in `PlayerEntity->onSImulation()` effectively doubled mouse sensitivity when `scene()->onSimulation()` was called twice (see issue #113). Since we've corrected the double-call, we also need to get rid of the `/2.0` in the mouse sensitivity.

Thanks to @pknowlesnv for pointing this out.

bboudaoud-nv:
This is a branch to finalize changes around mouse sensitivity in FPSci. In summary once we merge this branch we should have a good understanding of how mouse sensitivity works in FPSci and no longer require "fudge factors" to get a correct sensitivity value from DPI and cm/360 sensitivity.

This issues originates with Issue #113 (calling the scene `onSimulation()` twice) and shed light on multiple hard coded values that didn't need to be in the application.

Merging this request closes #120.

### DIFF
--- a/source/App.cpp
+++ b/source/App.cpp
@@ -98,9 +98,7 @@ void App::updateMouseSensitivity() {
     // G3D expects mouse sensitivity in radians
     // we're converting from mouseDPI and centimeters/360 which explains
     // the screen resolution (dots), cm->in factor (2.54) and 2PI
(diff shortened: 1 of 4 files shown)


bboudaoud-nv:
Before merging this pull request/closing issue #120 I suggest someone perform a quick "desktop" validation that we are correct using a mouse of known DPI/a constant translation distance to verify a full 360° turn at a few different cm/360 sensitivity values.

PR #122, opened 2020-04-24
Update for split 2D buffer/shaders

bboudaoud-nv:
This branch contains a first-pass attempt at splitting all 2D content into a separate buffer and further breaking out shaders to allow:

1. A shader to run specifically on 3D rendered content
2. A shader to run specifically on 2D (i.e. HUD-style) content
3. A shader to run on the composited 2D/3D content

All of these shaders are provided with "shadertoy-style" arguments. 

By splitting 2D (HUD) content from 3D rendering FPSci can also be used as a testbench for evaluation of techniques like split 2D/3D buffer compression and warp approaches that requrie this content to be seperated out.

Currently the weapon/view model is rendered into the 3D buffer, but could be drawn into a buffer of it's own. TODO: Create issue for this.

Additionally, this PR will create a separate control for render resolution(s) separate from the display resolution.


bboudaoud-nv:
We should think about whether we want to ability to include a split render of the view model here. It seems useful for features like late warp which might want to render that layer separately.

If we do I'm curious as to what the performance-based approach to this split-render would be. I'd presume we could set the view model visibility to `false` prior to calling `GApp::onGraphics3D()` to produce the "normal" scene render, then make our own call to `m_renderer->render()` with just the surfaces from the weapon model using the same lighting environment/render settings?

Is it okay for these two calls to share a single GBuffer? Can we use a frame buffer instead (i.e. we don't need depth testing between the weapon and the scene)? Do we want/need AO to occur between the view model and scene?

As it is we don't add/insert the weapon into the scene, but rather call the weapon's `onPose()` (which in turn calls `VisibleEntity->pose()` then `ArticulatedModel::pose()`) in `App:onPose()`. We could always pose the model into it's own surface array at this point, and produce a render from this.

bboudaoud-nv:
I believe we want to suggest the following model for resolution scaling in these new buffers:

1. If unspecified all resolutions default to the native render device/display sizing (1:1) w/ appropriate guard bands
2. We allow the 2D and 3D buffer to be scaled independently of render device buffer sizing (as an integer `Vector2` resolution in X/Y)
3. We allow the following options for when the 2D/3D buffer size doesn't match the display resolution:
     a. Do not scale, render a subset of the original picture and leave empty regions of a solid color at the edge
     b. Up/down scale using nearest neighbor or bilinear interpolation (which of these should we default to?)
4. These resolutions are set as part of the general config (i.e. can be session or experiment level) and are independent of the `windowSize` parameter used for non-fullscreen display from the startup config
5. When running in non-fullscreen mode the rendered resolutions still apply and G3D handles the final scaling to the new window size

pknowlesnv:
GL_NV_shading_rate_image and GL_NV_scissor_exclusive might be of interest to render lower resolution guardbands quickly. Even for a cube map, the shading rate could be scaled based on an angular distance from the forward vector. The output images would all be the same resolution but with less detail - I wonder if that could help compression much.

bboudaoud-nv:
I'm going to document the reason `split2DBuffer` exists here for now (rather than just relying on `shader2D` not being empty for the existence of this buffer).

As I see it currently the reason for this is 2 fold:

1. If you want to shade the 3D content w/o doing anything to the 2D content you need this buffer to avoid warping the 2D content w/ the 3D.
2. This might be useful for future solutions that would transfer this buffer, even if independent shading isn't required on the 2D content

Having said that we could remove `split2DBuffer` in favor of using a pass-through shader to create these buffers if we want to simplify the interface. I'm open to either approach.

jspjutNV:
It occurs to me that this change in render pipeline may have an impact on the runtime. We should try to do some profiling to see if it does (assuming cheap or no-op shaders, or even no shaders), and document the expected performance costs. 

The goal would be to have the `split2DBuffer = False;` case with no shaders set result in no change in performance, but allow other cases to have some performance hit that you pay for when you design that kind of experiment.

jspjutNV (approved):
Other than wanting some more detailed documentation (See #306 ), this appears to work reasonably well, and pass as many tests as things did before.

PR #124, opened 2020-04-24
Replace duplicate computation w/ updateMouseSensitivity()

bboudaoud-nv:
This adds a fix for a duplicated (incorrect) calculation from `App::updateSession()` with a call to `App::updateMouseSensitivity()`.


bboudaoud-nv:
We should test to make sure that updating the session (still) produces "correct" mouse sensitivity.

jspjutNV:
I tested this and it seemed correct both before, and after changing sessions within the margin of error of my testing method for 10, 20, 30 cmp360. I also tested 2 one time and it was pretty close (though cm were a high source of error, so smaller numbers were more error prone in my tests). I'm merging this request.

PR #132, opened 2020-05-11
Split SystemConfig into Input/Config Files and HW Information/Output Files

### LINKED ISSUE #128: SystemConfig Used for Output
Currently `systemconfig.Any` is used as a "pseudo-output" in that only the LDAT-R logging parameters are taken as input from this file. 

Specifically the following are _read from_ the config file:

* `HasLogger` - Does the system have an LDAT-R logger attached via VCP?
* `LoggerComPort` - If `HasLogger = `True` then provide the name of the LDAT-R logger COM port here
* `HasSync` - Does the system use an additional serial card signal for time-sync?
* `SyncComPort` - If `HasSync` = `True` then provide the name of the COM port used for sync signaling here

It may be worth considering renaming `systemconfig.Any` since all input/_true_ config fields used in this file are related to the LDAT-R/click-to-photon logger (currently).

The following parameters were purely informational/were never "read" from the `systemconfig.Any` by the application, but rather written back to this file on every run of the app (violating the _config_ naming convention):

* `CPU` - The CPU namestring (already provided in `log.txt` by G3D, though not w/ the same level of detail )
* `CoreCount` - The CPU core count (already provided in `log.txt` by G3D as `Num HW Threads`)
* `GPU` - The GPU namestring (already provided in `log.txt` by G3D w/ more detail)
* `MemoryCapacityMB` - The total system RAM (not provided in `log.txt`)
* `DisplayName` - Not currently working as expected and not provided in `log.txt`
* `DisplayResXpx` - The display (not window) horizontal resolution in pixels (not provided in `log.txt`)
* `DisplayResYpx` - The display (not window) vertical resolution in pixels (not provided in `log.txt`)
* `DisplaySizeXmm` - The display (physical) horizontal size in millimeters (not provided in `log.txt`)
* `DisplaySizeYmm` - The display (physical) vertical size in millimeters (not provided in `log.txt`)

Per @jspjutNV's suggestion I believe the parameters listed above should move to `log.txt`, the output database, or another (new) log file. They should _not_ be included in a file labelled as "config" as this is our convention for _input_.

bboudaoud-nv:
This branch splits the `SystemConfig` structure, previously used as an `inout` file into it's 2 constituent parts:

* System hardware configuration input (i.e. configuration for the LDAT-R logger interface)
* System hardware description output (CPU, GPU, RAM, Display Info)

Currently the system hardware info is (only) written to `log.txt` and duplicates some of the native information provided by G3D (see more about this in issue #128). We could instead/also:

* Write the system info to the output database
* Create a new output text/CSV file that only contains our (formatted) system info
* Use this information for automatic validation that the system meets some (specificed) min spec for certain experiments

Currently the `SystemInfo` class has been implemented (in `configFiles.h`) to get/store system hardware information separately from logger configuration parameters (using the existing `SystemConfig` class).

Merging this request closes #128.

### DIFF
--- a/source/App.cpp
+++ b/source/App.cpp
@@ -42,10 +42,13 @@ void App::onInit() {
 	userStatusTable.printToLog();
 	userStatusTable.validate(sessionIds);
 	
+	// Get info about the system
+	SystemInfo info = SystemInfo::get();
+	info.printToLog();										// Print system info to log.txt
+
 	// Get and save system configuration
 	systemConfig = SystemConfig::load();
-	systemConfig.printToLog();											// Print system info to log.txt
-	systemConfig.toAny().save("systemconfig.Any");						// Update the any file here (new system info to write)
+	systemConfig.printToLog();								// Print the system config to log.txt								
 
 	displayRes = OSWindow::primaryDisplaySize();						
 
--- a/source/App.cpp
+++ b/source/App.cpp
@@ -47,9 +47,10 @@ void App::onInit() {
 	info.printToLog();										// Print system info to log.txt
 
 	// Get and save system configuration
-	systemConfig = SystemConfig::load();
-	systemConfig.printToLog();								// Print the system config to log.txt								
+	latencyLoggerConfig = LatencyLoggerConfig::load();
+	latencyLoggerConfig.printToLog();						// Print the latency logger config to log.txt								
 
+	// Get the size of the primary display
 	displayRes = OSWindow::primaryDisplaySize();						
 
 	// Load the key binds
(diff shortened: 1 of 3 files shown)


bboudaoud-nv:
The division of input from output here as been finished. `LatencyLoggerConfig` handles input control, while `SystemInfo` handles logging the system information to file.

Currently all system info is logged to `log.txt`. If we'd like to add a table to the output database/write a new text/CSV file this is do-able, but I will wait to mark this pull request ready for review until I get feedback on what exactly we want as system info output.

Issue #131, opened 2020-04-29
Abstract `App` as `FPSciApp` to make merging into forks easier


jspjutNV:
There are some forks of FPSci that could benefit from easier merges from master if we allow them to specify their own `App` class that overrides whatever features from `FPSciApp` that they want. 

FPSci should still include a default `App` that will behave the same as the current one does.

This would be one step toward allowing FPSci to be abstracted into a library that would allow apps to use or ignore whichever features they want.

bboudaoud-nv:
One thing to discuss here is what exactly FPSci _is_/what this common base class should include.

* If we want to treat it as a starting point for (generic) FPS-style development then perhaps the `App` vs `Session` delineation should be revisited (i.e. `Session`-based approach could/should be optional). This would promote the idea of an `FPSApp` with generic FPS primitives in it.
* If we want to treat it as a scientific experimentation platform then the intertwining of `Session` and `App` responsibilities is potentially fine. This would be more in line with the idea of an `FPSciApp` base class that requires session-based implementation.

I (generally) find thinking about either the scientific control/measurement capabilities of FPSci as separate from the FPS-generic aspects of FPSci challenging right now, so this is likely something we can improve.

Generally speaking we could still benefit from a more modular approach to code design in FPSci and I think it makes sense to revisit this issue in a more thorough design document. Whether or not this level of effort makes sense is a functional of how useful/reusable we think the code base will be.

PR #134, opened 2020-05-24
Towards automated testing

pknowlesnv:
This sets up some basic test cases. They aren't all working yet. Targets are in the right position, but aren't rendered. Perhaps behind the wall or the camera isn't facing -z.

One thing to decide on is whether using multiple projects is a good idea for testing. The options are:

- Compile everything into the main app, but jumpt to testing via a --test flag. This is convenient being able to test the same binary. The drawback is extending test cases in other branches gets more complicated.
- Add the same source code to multiple projects. Everything's separate, but needs compiling twice.
- Compile a common static library with most of the code, then provide little executable projects that have entrypoints for either the mian app or testing. Switching between the projects to run tests is fast and convenient.

Merge request has some minor restructuring for use in other branches.


source/App.cpp
```diff
@@ -87,34 +86,42 @@ void App::onInit() {
...
+		// Set to first-person mode (i.e. hide the mouse cursor)
+		fpm->setMouseMode(FirstPersonManipulator::MouseMode::MOUSE_DIRECT);
+	}
 	// Control player motion using the experiment config parameter
 	shared_ptr<PlayerEntity> player = scene()->typedEntity<PlayerEntity>("player");
 	if (notNull(player)) {
-		player->mouseSensitivity = (float)mouseSensitivity;
-		player->turnScale = currentTurnScale();
+		player->m_pixelsToRadians = (float)pixelsToRadians;
+		player->turnScale = sessConfig->player.turnScale * userTable.getCurrentUser()->turnScale;
```
bboudaoud-nv, line 124: This is not a replacement for the `App::currentTurnScale()` logic, which includes scaling for when the player is in "scoped" mode. I'm not sure that is an issue given the context here (typically only when the player changes), but I suppose technically the user _could be_ in scoped mode when this happens (causing the loaded turn scale to be incorrect).

I assume this is okay here b/c we assume the user isn't scoped/call `currentTurnScale()` elsewhere, but is there a good reason to avoid calling that function here?

source/App.cpp
```diff
@@ -424,7 +438,7 @@ void App::updateParameters(int frameDelay, float frameRate) {
 	float dt = 0;
 	if (frameRate > 0) dt = 1.0f / frameRate;
 	else dt = 1.0f / float(window()->settings().refreshRate);
-	setFrameDuration(dt, GApp::REAL_TIME);
+	setFrameDuration(dt, m_simTimeStep);
```
bboudaoud-nv, line 441: This is another case of circular logic in calling  `GApp::setFrameDuration()`, which is the setter for `m_simTimeStep`. We want to set the frame duration using `REAL_TIME` here based on the parameters passed into the function. I can't think of a case where we would use anything of than a `REAL_TIME` so I think it makes sense to hard code that here or offer it up as an additional (optional) function parameter.

Unless there is a reason I'm missing I think inheriting the time-base here and the time somewhere else to make 2 calls to `GApp::setFrameDuration()` is more confusing than necessary.

FirstPersonScienceApp.vcxproj
```diff
@@ -0,0 +1,167 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bboudaoud-nv, line 1: Maybe we want to consider keeping `FirstPersonScience` as the application (rather than renamed to `FirstPersonScienceApp` and rename what is `FirstPersonScience` above to something like `FPSciLib` or something like that?

FirstPersonScience.vcxproj
```diff
@@ -40,13 +40,13 @@
     <CharacterSet>MultiByte</CharacterSet>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
-    <ConfigurationType>Application</ConfigurationType>
+    <ConfigurationType>StaticLibrary</ConfigurationType>
```
bboudaoud-nv, line 43: In the spirit of keeping older workflows working I suggest we rename this project to `FPSciLib` or something equivalent to keep the `FirstPersonScience` as an application and introduce a new project for the library.

source/App.cpp
```diff
@@ -87,34 +86,42 @@ void App::onInit() {
...
 
 /** Handle then user settings window visibility */
 void App::openUserSettingsWindow() {
-    m_userSettingsMode = true;
-    m_userSettingsWindow->setVisible(m_userSettingsMode);
+	m_userSettingsMode = true;
+	m_userSettingsWindow->setVisible(m_userSettingsMode);
+}
+
+/** Handle then user settings window visibility */
```
jspjutNV, line 99: Is "then" a typo for "the"? Also, I don't think this function is needed, particularly because I believe `m_userSettingsMode` is going away in #108 .

source/App.cpp
```diff
@@ -87,34 +86,42 @@ void App::onInit() {
...
-        fpm->setMouseMode(FirstPersonManipulator::MouseMode::MOUSE_DIRECT_RIGHT_BUTTON);
-    }
-    else {
-        // Set to first-person mode (i.e. hide the mouse cursor)
-        fpm->setMouseMode(FirstPersonManipulator::MouseMode::MOUSE_DIRECT);
-    }
+	const UserConfig* user = userTable.getCurrentUser();
+	// Converting from mouseDPI (dots/in) and sensitivity (cm/turn) into rad/dot which explains cm->in (2.54) and turn->rad (2*PI) factors
+	// rad/dot = rad/cm * cm/dot = 2PI / (cm/turn) * 2.54 / (dots/in) = (2.54 * 2PI)/ (DPI * cm/360)
+	const double pixelsToRadians = 2.0 * pi() * 2.54 / (user->cmp360 * user->mouseDPI);
```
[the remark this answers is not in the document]
jspjutNV: DPI is also not the most correct term, though it is commonly used. I believe the technically correct term is CPI (cycles per inch), but I've only seen a couple mouse manufacturers actually use that term, even though I believe many sensor companies use CPI in their spec sheets. `dotsToRadians` sounds like a function. Maybe `radiansPerDot` is a better name for the variable. Since it's related to the mouse input to camera control, maybe `cameraRadiansPerMouseDot` is the most verbose name for it.

source/PlayerEntity.cpp
```diff
@@ -114,7 +114,7 @@ void PlayerEntity::updateFromInput(UserInput* ui) {
 	m_jumpPressed = false;
 
 	// Get the mouse rotation here
-	Vector2 mouseRotate = ui->mouseDXY() * turnScale * (float)mouseSensitivity;
+	Vector2 mouseRotate = ui->mouseDXY() * turnScale * (float)m_pixelsToRadians;
```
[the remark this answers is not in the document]
jspjutNV: Reading this line where it's used, I think `turnRadiansPerMouseDot` or `cameraRadiansPerMouseDot` would make the variable name show that it's being used correctly here.

source/App.cpp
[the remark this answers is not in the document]
pknowlesnv: Yes, I much perfer the more explicit radiansPerDot - there's no ambiguity of an implied multiply.

pknowlesnv:
> Generally speaking I think we may want to create another tracking repo/fork for testing. 

I think it's important for testing to exist in the same repository, howeve we do it. Separating tests leads to potential versioning issues and extra effort to make sure one repo doesn't break the other (e.g. when a new test needs a new feature or when improving a feature breaks an existing test). It must be convenient to run tests, both before checking in code and for debugging failures. It's also important for the tests to be easily updated - there's less benefit if the time spent maintaining tests is greater than what you'd spend fixing/debugging issues that would occur otherwise.

jspjutNV:
This pull request would close #131 .

source/ConfigFiles.h
```diff
@@ -69,6 +69,11 @@ class StartupConfig {
     String userConfig() {
         return userConfigPath + "userconfig.Any";
     }
+	
+    /** filename with given path to user status file */
+	String userStatusConfig() {
+		return userConfigPath + "userstatus.Any";
```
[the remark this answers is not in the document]
pknowlesnv: I.e. replace `userConfigPath` with `userConfigFile`?

This changes is needed to support testing with separate configs. Most of the tests need to be run in a clean environment and it would be a pain if every little change made to the main experiment configs kept breaking the tests.

I made a single separate test case to validate the main configs allow the app to run.

source/PlayerEntity.h
```diff
@@ -126,6 +126,10 @@ class PlayerEntity : public VisibleEntity {
 
 	void respawn() {
 		m_frame.translation = m_respawnPosition;
+		m_headingRadians = 0.0f;
```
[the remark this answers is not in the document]
pknowlesnv: The problem with this is respawning the player would not reset the heading to some absolute value, which breaks testing. I'd be happy if this reset the player to the value in the config. Maybe respawn() should be handled outside the player class, or it needs to keep the respawn CFrame?

source/App.cpp
```diff
@@ -87,34 +86,42 @@ void App::onInit() {
 	updateMouseSensitivity();			// Update (apply) mouse sensitivity
 	updateSessionDropDown();			// Update the session drop down to remove already completed sessions
 	updateSessionPress();				// Update session to create results file/start collection
+
+	setFrameDuration(m_wallClockTargetDuration, REAL_TIME);
```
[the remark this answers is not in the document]
pknowlesnv: Yes, this is just needed to set GApp::REAL_TIME. I don't want to touch the initial m_wallClockTargetDuration value but G3D doens't have any way to do this individually. I removed the setFrameDuration call from updateParameters() in the second patch because it breaks testing, which uses a fixed timestep for frame perfect tests.

source/App.cpp
pknowlesnv: oops. looks like this is an integration mistake. should only be changing the first line. thanks!

source/App.cpp
pknowlesnv: I'd argue it's better to set a value only when it needs to change rather than hammering it every frame. The reason here is the test wants to set GApp::MATCH_REAL_TIME_TARGET for a fixed timestep and needs this not to change every time it calls oneFrame().

FirstPersonScienceApp.vcxproj
pknowlesnv: Sounds good to me. Should we keep everything `FirstPersonScience` or `FPSci` as well?

source/App.cpp
pknowlesnv: Looks like it. I can change it. The tests will need some way to close the settings window, but it could inject Esc just as well.

jspjutNV: I believe you can call `m_userSettingsWindow->setVisible(false);` to hide the window. Might still need the `m_userSettingsMode = false;` in the current version, though that line can go away once the other PR is merged.

If that's not doable, then I can see why you might need the callback function and how that would be cleaner for the test.

bboudaoud-nv: Technically speaking we don't really need `m_userSettingsMode` we can just use the visible flag (`m_userSettingsWindow->visible()`) for the window to manage whether we are in this mode.

For example `m_userSettingsWindow->setVisible(!m_userSettingsWindow->visible())` would suffice.

pknowlesnv:
I've made the requested updates and also renamed App as per #131. Unfortunately `master` at 3bbeef9dd92ec770aaf23040729b408a7a6873c3 is asserting for me on startup with the following:

>	FirstPersonScience.exe!G3D::G3DString<64>::testInvariants() Line 122	C++
 	FirstPersonScience.exe!G3D::G3DString<64>::maybeDeallocate() Line 244	C++
 	FirstPersonScience.exe!G3D::G3DString<64>::operator=(const G3D::G3DString<64> & s) Line 435	C++
 	[External Code]	
 	FirstPersonScience.exe!UserMenu::setConfig(MenuConfig & config) Line 178	C++
 	FirstPersonScience.exe!App::updateControls() Line 209	C++
 	FirstPersonScience.exe!App::updateSession(const G3D::G3DString<64> & id) Line 353	C++
 	FirstPersonScience.exe!App::onInit() Line 93	C++

source/PlayerEntity.h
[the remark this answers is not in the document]
pknowlesnv: I added another variable to record the initial value. Ideally we'd have an euler rotation yaw/pitch/roll struct to hold everything in one go.

source/App.cpp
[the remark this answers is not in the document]
pknowlesnv: I added some comments to these areas to clarify.

FirstPersonScience.vcxproj
pknowlesnv: done

source/ConfigFiles.h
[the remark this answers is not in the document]
bboudaoud-nv: I may have been confused here. I think the "right" thing to do is let the experiment designer specify a full path for the user status/config independently without any file system tie between the two (i.e. co-location in a single directory). This way we could name our test files `test_userstatus.Any` and not conflict with a default `userstatus.Any` file.

An example motivating case would be something like a set of experiments hosted through an NFS where the designer might want a `userConfigs` directory and a `userStatus` directory each containing multiple user config/status files.

If this doesn't apply here, we can always take the changes as-is and create a new issue to implement this as a standalone feature request.

jspjutNV:
> I've made the requested updates and also renamed App as per #131. Unfortunately `master` at [3bbeef9](https://github.com/NVlabs/abstract-fps/commit/3bbeef9dd92ec770aaf23040729b408a7a6873c3) is asserting for me on startup with the following:
> 
> > FirstPersonScience.exe!G3D::G3DString<64>::testInvariants() Line 122	C++
> > FirstPersonScience.exe!G3D::G3DString<64>::maybeDeallocate() Line 244	C++
> > FirstPersonScience.exe!G3D::G3DString<64>::operator=(const G3D::G3DString<64> & s) Line 435	C++
> > [External Code]
> > FirstPersonScience.exe!UserMenu::setConfig(MenuConfig & config) Line 178	C++
> > FirstPersonScience.exe!App::updateControls() Line 209	C++
> > FirstPersonScience.exe!App::updateSession(const G3D::G3DString<64> & id) Line 353	C++
> > FirstPersonScience.exe!App::onInit() Line 93	C++

Now that #140 has been merged into master, we should try this one again to see if the startup bug has been fixed.

pknowlesnv:
Rebased and updated with the new project names.

I also updated the build directory to better separate each project.

data-files/test/experimentconfig.Any
```diff
@@ -0,0 +1,93 @@
+{
```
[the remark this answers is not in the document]
jspjutNV: These appear to be in the `test` directory inside `data-files`. I'd prefer for them to be somewhere else, though it's probably not a big problem for them to be there since they could be used as reference configs in a pinch. I'll resolve this conversation since this location is better then being in the root of `data-files`.

source/PlayerEntity.cpp
```diff
@@ -54,7 +54,7 @@ shared_ptr<Entity> PlayerEntity::create
 void PlayerEntity::init(AnyTableReader& propertyTable) {
    // Get values from Any
     Vector3 v;
-	float heading = m_headingRadians;
+	float heading = 0.0f;
```
jspjutNV, line 57: @bboudaoud-nv , this isn't a bug is it? I know there have been issues with getting the heading to load correctly from configs before.

tests/main.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+	// Stop visual studio creating an abort popup and stalling the CI runner
+	_CrtSetReportHook(AbortReportHook);
+
+	// Hack to disable error popups during automated testing because there is no dynamic version of G3D_DEBUG_NOGUI
+	G3D::_internal::_consolePrintHook = nullptr;
+	G3D::_internal::_debugHook = nullptr;
+	G3D::_internal::_failureHook = nullptr;
+
+	::testing::InitGoogleTest(&argc, const_cast<char**>(argv));
```
jspjutNV, line 35: VS is giving red underlines here. Is something missing to make this seem to work more correctly to visual studio?

tests/main.cpp
```diff
@@ -0,0 +1,56 @@
+
```
jspjutNV, line 1: I get the following error from all of the `tests/*.cpp` files:
> 1>C:\Users\jbspj\Documents\GitHub\NVIDIA\abstract-fps\tests\main.cpp(57,1): fatal error C1010: unexpected end of file while looking for precompiled header. Did you forget to add '#include "pch.h"' to your source?

tests/main.cpp
pknowlesnv: Perhaps the const cast? If it compiles I'd claim an intellisense bug and call it a day. Personally I think argv should be const, and googletest could happily take a const anyway, but main typically has a non-const argv.

tests/main.cpp
pknowlesnv: That's a problem. I thought I disabled precompiled headers. Maybe I forgot to change the dropdown box to release _and_ debug rather than just debug. Will update.

tests/main.cpp
jspjutNV: It's complaining about the `::` with nothing in front of them. Shouldn't there normally be a namespace or something on the left hand side of the `::` operator?

pknowlesnv: `::` alone is valid C++. It's an explicit way to refer to the global namespace in case a local/class member has the same name. I doubt it's a new feature either. This line is also from the official googletest docs: https://github.com/google/googletest/blob/master/googletest/docs/primer.md

Nevertheless I've removed it. Hopefully it's happy now?

pknowlesnv:
Rebased on top of master again. There were some conflicts and a crash from FirstPersonManipulator using the old UserInput during the tests. The number of passing tests has dropped from 5 to 4 too.

jspjutNV:
It looks like this is building for me now, and I get the same 4 tests passed and 8 failed. I'll do one last pass over this set of changes, then I'll merge it in if everything looks okay and we can start to fix the tests that are failing. Thanks a lot for doing this!

jspjutNV:
I just noticed a bug in this where pressing any movement key forces the heading to a single direction during the button press. The mouse heading appears to continue to update while the movement happens and as soon as movement ends, the mouse heading regains control.

jspjutNV:
I fixed the bug. I think this is ready to merge. Thanks again Py and great job!

PR #137, opened 2020-05-27
User save settings

### LINKED ISSUE #136: User settings save button is missing
The `allowUserSettingsSave` flag should be serialized from the experiment config.

Also the save button should show up regardless of the `allowReticleChange` setting (currently they appear dependent).

bboudaoud-nv:
This pull request handles 2 problems with the user settings save option:

1. It was not previously serialized to/from Any
2. It was drawn conditional on the `allowReticleChange` flag, which is incorrect

Merging this request closes #136.


jspjutNV:
In fixing the documentation for this change, I noticed an inconsistency in the `reticleShrinkTime` variables so I renamed them all to be `reticleChangeTime` since that's the more accurate term. I believe this is ready to merge.

PR #152, opened 2020-06-11
Run commands

### LINKED ISSUE #148: Run arbitrary commands from FPSci
Currently the FPSci solution is hard-coded to run the research latency logger platform around start of sessions. We have a request to support running others scripts/programs instead. It makes sense to consider adding fully configurable command calls around FPSci runtime for diverse experimenter needs. 

This would/could include running arbitrary commands before/after the start of an application, session, block, or even trial. We should discuss which fields are useful and consider adding those fields which would be most helpful.

bboudaoud-nv:
This branch develops features around allowing the user to specify commands/scripts that are run at the start/end of a once per app, session, and trial basis.

Merging this PR closes #148.


jspjutNV:
I tested this branch with the following session in my config and it appears to work as expected, except I notice that in the current version, the second commands in the lists don't appear to be executed.
```
        {
            description = "Test Commands";
            frameRate = 240;
            id = "commands";
            trials = ( {count = 2; ids = ( "static" );}, {count = 2; ids = ("static", "static") } );
            commandsOnSessionStart = ( "cmd /c echo Session start>> commandLog.txt", "cmd /c echo Session start second command>> commandLog.txt" );
            commandsOnSessionEnd = ( "cmd /c echo Session end>> commandLog.txt", "cmd /c echo Session end second command>> commandLog.txt" );
            commandsOnTrialStart = ( "cmd /c echo Trial start>> commandLog.txt" );
            commandsOnTrialEnd = ( "cmd /c echo Trial end>> commandLog.txt" );
        }
```

jspjutNV:
I recommend we open an issue for the remaining bug, and merge this in so we can begin the next release.

PR #153, opened 2020-06-11
Automatic session progression

### LINKED ISSUE #151: Automatically Progress Sessions
The current implementation brings up the user settings menu after the completion of each session. This behavior is useful, but isn't what we'd like for an experiment that's ready to run. Let's add an experiment config setting to prevent the user settings menu from showing between sessions.

bboudaoud-nv:
This branch implements flags for control over whether:

* The user menu is open at app startup (`showMenuOnStartup`)
* The user menu is opened automatically at the end of each session (`showMenuBetweenSessions`)

Merging this pull request closes #151


bboudaoud-nv:
The only (potential) remaining item here is that I don't believe sessions print a message to the screen at their start. This isn't an issue/PR can be closed if the behavior is fine as-is.

PR #155, opened 2020-06-11
FPSci packager generator

jspjutNV:
This is a script that consumes a `log.txt` to generate a 1-time use distribution generating script that will put all the required binaries in a single `dist` directory that can be used to distribute FPSci to people without Visual Studio.


.gitignore
```diff
@@ -30,3 +30,6 @@ data-files/model
 packaged
```
bboudaoud-nv, line 30: The `packaged` directory was my (old) version of what now appears to be `dist`. Since we are moving to a different name we can probably drop this from the `.gitignore` (and I can rename that default directory in my packager script if we prefer `dist`).

scripts/package/fpsci_packager.py
```diff
@@ -0,0 +1,68 @@
...
+# Name of the packaging script to generate
+outputScript = 'package.sh'
+
+# Path of the log.txt file to use
+inputLog = 'data-files/log.txt'
+
+# The list of files that should always be excluded
+excludeList = ['log.txt', 'keymap.Any', 'systemconfig.Any']
+# The list of files that should be excluded for release builds, but may want to be included for specific experiment distributions
+genericExcludeList = ['experimentconfig.Any', 'startupconfig.Any', 'userconfig.Any', 'userstatus.Any']
```
bboudaoud-nv, line 14: Could use a more descriptive name here, something like `experimentSpecificExcludeList`, but this is largely preference.

scripts/package/fpsci_packager.py
```diff
@@ -0,0 +1,68 @@
...
+
+# List of paths to search for to merge into the same distribution directory
+basePath = ['data-files', 'game', 'common']
```
bboudaoud-nv, line 18: In the `g3d_packager.py` script I also included `research` here, but we may not be using anything from that sub-directory.

scripts/package/fpsci_packager.py
```diff
@@ -0,0 +1,68 @@
...
+
+# Distribution path - where the output script will generate the distribution
+distPath = 'dist/'
```
bboudaoud-nv, line 21: Could be nice to make this an (optional) command line argument, but not required.

scripts/package/fpsci_packager.py
```diff
@@ -0,0 +1,68 @@
+# This script generates a packaging script based on the files used by the last successful run of the program.
+# You need to run the generated script to create the distribution directory, then zip that up to send to people.
+
+# Name of the packaging script to generate
+outputScript = 'package.sh'
+
+# Path of the log.txt file to use
+inputLog = 'data-files/log.txt'
```
bboudaoud-nv, line 8: The input log file and output script name could be (optional) command arguments, but not required.

scripts/package/fpsci_packager.py
```diff
@@ -0,0 +1,68 @@
...
+        # In the file list, add this file to what we want to copy
+        if not beforeFiles and line != '\n':
+            filename = line.strip()
+            basename = filename.split('/')[-1]
+            # Exclude the lists of files we don't want to package
+            if basename in excludeList or basename in genericExcludeList:
+                continue
+            
+            # Absolute path given
+            if filename.startswith('C:/') or filename.startswith('c:/'):
```
bboudaoud-nv, line 60: This may create issues for folks packaging from a build in a drive with a different name. Can go as-is and be updated later. Could use that the 2nd-3rd character is `:/` to detect absolute paths.

.gitignore
jspjutNV: I really don't know what it should be called... I've gone back and forth on naming, and you can see my script is called "packager". Since the testing infrastructure uses `packages` for the test stuff, I figured having a different name made sense. 

I propose we leave `packages` for external things we're bringing in and `dist` for stuff we're intending to distribute to others.

scripts/package/fpsci_packager.py
jspjutNV: What about `secondaryExcludeList`? 

It's hard to name this thing because it's only here to plan for when we have 2 different types of builds fully supported, one for the generic FPSci and the other for a purpose-built FPSci-based experiment. Calling it `experimentSpecific` makes it seem like you want that list when you're building an experiment, but I think we remove this exclude list for a specific experiment...

scripts/package/fpsci_packager.py
jspjutNV: I don't think FPSci is using anything from the `research` directory, but it may be needed in the future. Something we should keep an eye on.

scripts/package/fpsci_packager.py
jspjutNV: Yeah, let's add command line stuff in a future version.

scripts/package/fpsci_packager.py
jspjutNV: Good suggestion, let's do that.

I don't like being so windows-specific, but that's just where we are right now.

scripts/package/fpsci_packager.py
bboudaoud-nv: Hmm, I think I may have misinterpreted it's use. If the intent is to exclude this content for "generic" builds I can now see the descriptive nature of the naming. Maybe we leave it as is.

scripts/package/fpsci_packager.py
jspjutNV: Let's make command line options a future addition.

PR #168, opened 2020-06-15
Experiment designer control of whether to show reference target

### LINKED ISSUE #164: Flag for controlling reference target
Per request from @joohwankimNV add a general config control flag to allow the experiment designer to configure whether the reference target is drawn within an experiment/session.

bboudaoud-nv:
This branch implements the `showRefTarget` flag for session/experiment level configuration that allows the user to specify that the reference target (not) be drawn between trials within the session.

Merging this PR closes #164.

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -298,6 +298,7 @@ These flags help control the behavior of click-to-photon monitoring in applicati
 | Parameter Name        |Units                  | Description                                                                        |
 |-----------------------|-----------------------|------------------------------------------------------------------------------------|
 |`targetHealthColors`   |[`Color3`, `Color3`]   | The max/min health colors for the target as an array of [`max color`, `min color`], if you do not want the target to change color as its health drops, set these values both to the same color                                              |
+|`showRefTarget`        |`bool`                 | Show a reference target to re-center the view between trials/sessions?             |
 |`referenceTargetColor` |`Color3`               | The color of the "reference" targets spawned between trials                        |
 |`referenceTargetSize`  |m                      | The size of the "reference" targets spawned between trials                         |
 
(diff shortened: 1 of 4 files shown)


docs/general_config.md
```diff
@@ -298,6 +298,7 @@ These flags help control the behavior of click-to-photon monitoring in applicati
 | Parameter Name        |Units                  | Description                                                                        |
 |-----------------------|-----------------------|------------------------------------------------------------------------------------|
 |`targetHealthColors`   |[`Color3`, `Color3`]   | The max/min health colors for the target as an array of [`max color`, `min color`], if you do not want the target to change color as its health drops, set these values both to the same color                                              |
+|`showRefTarget`        |`bool`                 | Show a reference target to re-center the view between trials/sessions?             |
```
jspjutNV, line 301: Given that the other parameters are `referenceTargetColor` and `referenceTargetSize`, I believe this feature would work better being called `showReferenceTarget`.

source/ConfigFiles.h
```diff
@@ -1323,6 +1323,7 @@ class TargetViewConfig {
 	float           combatTextTimeout = 0.5f;							///< Time for combat text to disappear (in seconds)
 
 	// Reference/dummy target
+	bool			showRefTarget = true;								///< Show the reference target?
```
jspjutNV, line 1326: Even though I asked to change the name of the config parameter, it's fine to leave this as `showRefTarget` because it matches the other C++ names for reference target variables.

jspjutNV:
I fixed the changes I requested, but this branch has a bug where after completing a question response, the mouse cursor remains visible while it begins controlling the aim again. This bug should be fixed before merging this pull request.

bboudaoud-nv:
@jspjutNV is this persistent or does it only last for a moment? In trying to reproduce this error I have no been able to keep the mouse on screen between sessions/following the questions being displayed as you've described. Some steps to reproduce the error would be great.

bboudaoud-nv:
In addition, I notice that we no longer support the scoreboard duration, which would allow the experiment developer to time the duration over which the session complete message is shown prior to moving on to a new session. This seems more important when the reference target is not shown.

Do we want to consider adding back the scoreboard duration as part of this branch/pull request to make this configurable?

jspjutNV:
> Do we want to consider adding back the scoreboard duration as part of this branch/pull request to make this configurable?

I think we do want that. We'd like to be able to control the duration of the experiment, and also control the timing for the session complete message and session begin message (I forget if that feature exists).

jspjutNV:
I think this is ready to review, though it seems to have a conflict with master that will need to be resolved

PR #173, opened 2020-06-18
Move currentUser from user config to user status

### LINKED ISSUE #170: Move current user field to user status file
Creating an issue to track the idea @jspjutNV had about moving the `currentUser` field to the user status file (from the user config). 

Doing this would eliminate the need to have an additional write-back to the user config file when the current user is changed at runtime, as we are already writing back to the user status regularly (once per session) to update the completed sessions array(s). This would also separate the "pure input" file (user config) from the regularly written-back status.

bboudaoud-nv:
This branch moves the `currentUser` configuration parameter from the user config file to the user status file (to be more consistent with the rest of the application's use of these files).

Including @pknowlesnv as a reviewer since this prompted a small change in the `FPSci.test` project.

Merging this PR closes #170.


jspjutNV:
I got the tests passing and improved the error message for old config file versions. This one is ready to merge.

Issue #133, opened 2020-05-12
Named Spawn Locations in Scene Files


bboudaoud-nv:
@pknowlesnv suggested that we supported named spawn locations in our `scene.any` file specification. This would allow the experiment/session level configuration to specify different spawn points within a scene. This would be supported as either:

1. Part of our `PhysicsScene` class
2. Part of a child class to `PhysicsScene` that specifies more FPSci-specific behavior

Player "respawn" behavior can be set to either use a fixed spawn point or revert to the closest spawn point to where the player ended the last session/fell below the respawn height.

If no spawn point is specified in the experiment config then the first "named" location is used as the spawn location. These spawn locations could/should include both the player position (XYZ) as well as view direction (heading).

bboudaoud-nv:
@morgan3d suggested we use `MarkerEntity`s to setup the spawn points in the scene in a G3D compliant way. This makes sense and is something we should look into as a solution that avoids duplicating effort.

We would need to add a `String` (or possibly `Array<String>` if we want "randomized" spawn behavior) field to the `FpsConfig` to allow the experiment designer to specify a named marker entity from the scene file at which to spawn the player.

jspjutNV:
I like both of those recommendations. I think the `MarkerEntity` one is high priority and we should get done as soon as reasonable since it affects workflow for designing new scenes and experiments generally.

The second one for randomizing spawn behavior by listing the spawn locations in the `FpsConfig` could probably wait until someone has a need for it. It may also run into issues where a set of targets needs to be matched with the spawn location, though I suppose that can be partially solved by using separate sessions for coupled spawns and targets.

bboudaoud-nv:
This could be implemented as a part of the longer term restructure suggested in #174.

jspjutNV:
@bboudaoud-nv I believe #176 addressed this issue and we can close it. I'm closing this, but if it turns out that we needed something else for this, we can re-open.

For what it's worth, I believe this is solved by specifying different cameras in the scene and then specifying which camera to use in the scene structure per experiment/session as wanted.

PR #176, opened 2020-06-25
Scene configuration

### LINKED ISSUE #174: FPSci-specific scene configuration
Currently FPSci `scene.any` files require additional fields not found in "default" G3D scenes. These include:
* The `Physics` descriptor field (borrowed from the `simpleGame` sample's `PhysicsScene` implementation
* The `player` descriptor field (used to set spawn location/heading for our `PlayerEntity` class)
* A `playerCamera` camera descriptor field, used to defined the camera that is (implicitly) linked to the `playerEntity`

Additionally, current/future feature requests tend towards further increasing information that might be useful to store in the `scene.any` file including: 
* Issue #133 has a request for adding named spawn locations to the scene file
* The possibility of storing (named) world-space targets within a scene file (then optionally referencing them from an experiment config) so that the target motion is consistent with scene geometry

In an effort to allow maximum flexibility of the application and re-use G3D's existing `scene.any` files it makes sense to consider a way to load/reuse default G3D scenes, but also customize them without a need to edit the original scene files. Perhaps we should consider an `fpsci.scene.any` file format affiliated with the G3D `scene.any` file explicitly/by name that would allow us to add this data to the scene without a need for modifying/appending to the scene file.

bboudaoud-nv:
This branch adds functionality for a `scene` parameter within the general configuration structure (and a corresponding `SceneConfig` class within our configuration heirarchy). This allows the user to take "clean"/unmodified G3D scene files and add FPSci-specific configuration to them in an experiment config file (rather than in the scene file itself).

New `.fpsci.scene.Any` files can be created and included as the scene using `scene = #include("myfile.fpsci.scene.Any")` at an experiment/session level.

Merging this PR closes #174.


jspjutNV:
Looking at this pull request, I believe some documentation updates will be needed to reflect this change.

bboudaoud-nv:
Currently the `SceneConfig` supports:
* Scene name (same as historically)
* Player camera name override (defaults to the scene's `defaultCamera`)
* Reset height override (defaults to the scene's `Physics` field's `minHeight` parameter if specified)
* Spawn position/heading override (defaults to the scene's `PlayerEntity` settings if specified)

If there are additional fields we'd like to support (i.e. `MarkerEntity` based spawn locations) we can add these as well, but we'll need a full description of how this functionality is intended to work within the application.

bboudaoud-nv:
I'm going to open this PR for review. If there are any additional requests for fields that should be specifiable in the Scene config we can add them in the PR during the review/as new issues.

jspjutNV:
This pull request appears to break compatibility with previous experiment configs. @joohwankimNV , are you prepared to accept this change?

jspjutNV:
> This pull request appears to break compatibility with previous experiment configs. @joohwankimNV , are you prepared to accept this change?

@joohwankimNV told me that this change is a welcome one. I'm working on testing this change an improving error messaging accordingly.

jspjutNV:
Will this one also close #133 ?

jspjutNV:
8 of the tests currently fail. I'll need to investigate this failure, and likely the test configs will need to be updated to work correctly.

jspjutNV:
Good news, after merging in master the target sizes test also passes, so we have 5 passing and the same 8 failing.

jspjutNV:
So far everything I've tried to do that appears to match the documentation has caused the tests to crash. I'm not sure what I'm doing wrong. I've tried:

```
    scene = {
        name = "G3D Simple Cornell Box (Empty CO)";
    }
```
and a few similar variants.

jspjutNV:
I've discovered what I believe to be a bug due to `FPSci::updatePlayer()` and `FPSci::updateFromSceneConfig(player)` having overlapping functionality. A fix is in progress.

jspjutNV:
I believe this one is finally ready to merge. I improved the documentation, merged a couple functions that were fighting, and made the player spawn at the specified location on start of each session as well as at first launch. All tests are passing as well with the updated scene in the experiment config.

PR #181, opened 2020-06-25
Fix conditional ordering to prevent crash with no valid sessions

jspjutNV:
This corrects a regression where the app would crash if started with no valid sessions available for the user.

### DIFF
--- a/source/Session.cpp
+++ b/source/Session.cpp
@@ -322,14 +322,14 @@ void Session::updatePresentationState()
 		}
 	}
 	else if (currentState == PresentationState::scoreboard) {
-		if (stateElapsedTime > m_config->timing.scoreboardDuration) {
-			newState = PresentationState::complete;
-			if (m_hasSession) {
+		if (m_hasSession) {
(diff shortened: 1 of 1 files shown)


source/Session.cpp
```diff
@@ -322,14 +322,14 @@ void Session::updatePresentationState()
 		}
 	}
 	else if (currentState == PresentationState::scoreboard) {
-		if (stateElapsedTime > m_config->timing.scoreboardDuration) {
-			newState = PresentationState::complete;
-			if (m_hasSession) {
+		if (m_hasSession) {
+			if (stateElapsedTime > m_config->timing.scoreboardDuration) {
+				newState = PresentationState::complete;
```
bboudaoud-nv, line 327: Previously the new state was set unconditional of `m_hasSession` do we want to keep that? Seems like this should be above line 325.

jspjutNV: I want the state progression to be gated by the timing... but checking the timing needs to be stopped if there's no session set. Seems like we'd need another level of nesting to satisfy all of these... I'll give that a try.

PR #184, opened 2020-06-29
Custom feedback

### LINKED ISSUE #182: Customizable feedback messages
@joohwankimNV has made a request for customizable feedback messages (as part of the general config) as opposed to the hard-coded messages we have now.

Options for message strings include:
* Initial message (shown prior to first reference target)
* Success/failure trial response messages (shown at the end of each trial)
* Block complete message
* Session complete message
* All sessions complete message

We should also consider options for allowing the user to format parameters (such as the current session name, block number, trial index, and trial score) into some/all of these messages.

bboudaoud-nv:
This branch adds support for customizable feedback messages based on experiment configuration. These messages include replaceable sub-strings that allow users to format runtime data from the application (such as session score, or trial time) into the messages.

The currently supported found-and-replaced sub-string list includes:

| Sub-string            | Description                                           |
|-------------------|-------------------------------------------------------|
|`%score`              | The score from the current session                        |
|`%lastblock`        | The index of the last block completed                   |
|`%currblock`       | The index of the current block                                |
|`%tasktimems`    | The time the previous task took to complete         |

We can add additional sub-strings if desired.

Merging this request closes #182.


bboudaoud-nv:
If there are any requests for additional sub-strings we should support for replacement, feel free to make them here/in the original issue.

jspjutNV:
Just looking at the name `%score` it's not immediately apparent to me what that means. 

Looking at the code, I believe the name `%totalTimeRemaining` would be more descriptive of what is actually printed. 

My thinking here is that the concept of "score" is something that the experiment designer may want to change depending on the design and that some of these options could potentially be used as score, or maybe other options we add in the future.

Other possible names for the same field: `%timeLeft`, `%timeleft`, `%totaltimeleft`, `%timelefts`, `%timeLeftS`.

Other sub-strings that could be worth adding: `%targetsHit`, `%totalTargets`, `%shotsHit`, `%totalShots`.

bboudaoud-nv:
I agree w/ most of the additiona sub-strings suggested by @jspjutNV. To make these a bit clearer I will adopt the suggested named change of  `%score` → `%totalTimeLeftS` (as suggested). This will be for a full session (as implied by "total").

In addition I will apply camel-casing consistently across all sub-strings (i.e. `%lastblock` → `%lastBlock`) and rename `%tasktimems` → `%trialTaskTimeMs` to indicate this time is affiliated with a single trial.

The following are quite easy to implement on a per-trial basis (for trial feedback messages). They are currently not stored in aggregate for the entire session. For now I've renamed them to imply they provide a single trial worth of feedback, but we can consider also adding per-session aggregates as well:
* `%targetsHit` → `%trialTargetsDestroyed`
* `%totalTargets` → `%trialTotalTargets`
* `%shotsHit` → `%trialShotsHit`
* `%totalShots` → `%trialTotalShots`

jspjutNV:
I tested with the following in my experiment config:

```
    // Testing custom feedback messages
    referenceTargetInitialFeedback: "Click to spawn a target, then use shift while aiming at red target to begin.";
    noReferenceTargetInitialFeedback: "Click to start the session with block %currBlock!";
    trialSuccessFeedback: "%trialTaskTimeMs ms! You hit %trialTargetsDestroyed/%trialTotalTargets targets. %totalTimeLeftS s left total.";
    trialFailureFeedback: "Failure! You hit %trialTargetsDestroyed/%trialTotalTargets targets.";
    blockCompleteFeedback: "Block %lastBlock complete! Starting block %currBlock/%totalBlocks.";
    sessionCompleteFeedback: "Session complete! You scored %totalTimeLeftS!";
    allSessionsCompleteFeedback: "All Sessions Complete!";

    taskDuration = 5;
```

Due to #185 , the session also needs to have `description = "training";` at the moment, but that should be fixed once #185 is resolved.

With my last crash fix, I believe this pull request is ready to merge.

PR #193, opened 2020-07-01
Quick fix for showExperimentSettings = false

### LINKED ISSUE #192: Exception when showExperimentSettings = false
When `showExperimentSettings = false` is in the experiment config the application crashes (due to the session drop down not having been created).

bboudaoud-nv:
This branch adds a fix for an exception that previously occurred when `showExperimentSettings = false` was added to the experiment config Any file.

Merging this PR closes #192.


bboudaoud-nv:
This approach just "hides" the elements of the menu rather than attempting to duplicate the function they play in other member variables/provide it independent of whether the experiment settings are shown. 

If we'd like I can move to a more complicated solution that doesn't hide these elements, but instead truly doesn't add them and tracks their values internally.

jspjutNV:
This looks good and seems to prevent the crash described. Merging

PR #195, opened 2020-07-01
Configure results directory path from startup config

### LINKED ISSUE #190: Release builds create `results` directory at the same level as FPSci
It may be that we intend for this to happen, but instead of creating a `results` directory inside of the FPSci directory, FPSci creates a `results` directory parallel to the FPSci directory on release builds. I'm sure users can work around it, and I understand why it happens. Is this what we want to have happen or should we change the behavior?

bboudaoud-nv:
This branch adds support for specifying a `resultsDirPath` in the startup config file to allow the experiment designer to redirect the results directory wherever they would like. It also adds support for creating an arbitrarily named results directory if one does not already exist.

Merging this PR closes #190

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ data-files/log.txt
 data-files/*.db
 data-files/*.csv
 data-files/*.pdn
-data-files/result_data
 data-files/*.Any
 data-files/*.jpg
 *.mp4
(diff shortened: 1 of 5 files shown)


jspjutNV:
I'm testing that the logger functionality isn't lost by putting the following at line 53 in `PythonLogger.cpp`:

```
		String cmd = "cmd /c echo Timestamp [s],Event>>\"" + logName + "_event.csv\" & echo 2019-04-22 16:37:53,SW sync>>\"" + logName + "_event.csv\"";
		cmd += " & echo 1484.463512,SW>>\"" + logName + "_event.csv\" & echo 1487.81404,M1>>\"" + logName + "_event.csv\" & echo 1487.93696,PD>>\"" + logName + "_event.csv\"";
```

In addition, the `systemconfig.Any` file needs to be generated with the following contents:
```
{
    HasLogger = true;
    HasSync = false;
}
```

jspjutNV:
I've tested relative and absolute paths and I believe this is working as intended.

PR #199, opened 2020-07-02
Target previews

### LINKED ISSUE #187: Target preview during reference target display
@joohwankimNV has requested the ability to "preview" a trial's target(s) while the reference target is being shown. These targets cannot be damaged until after the reference target is destroyed, but allow the user to become familiar with target motion ahead of starting the timed trial.

@jspjutNV has suggested we add two flags (to the session config?) to support this:
* `spawnInactiveWithReferenceTarget` which, when true, indicates that all of a trial's targets should spawn with the reference target, but not be able to be hit until after that reference target is destroyed
* `inactiveTargetColor` a color field that allows the user to set the color of inactive targets as a way to demonstrate that they are not currently hit-able

One or both of these fields may be implemented based on discussion in this issue.

bboudaoud-nv:
This branch adds support for showing a user "previews" of real targets while the reference target is being drawn. The feature is enabled using the `showPreviewTargetsWithReference` flag in a session/experiment config.

The preview targets can be delineated from "real" targets using their color (settable with the `previewTargetColor` parameter in the session/experiment config).

Merging this PR closes #187.

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -344,7 +344,8 @@ These flags help control the behavior of click-to-photon monitoring in applicati
 |`showReferenceTarget`   |`bool`                | Show a reference target to re-center the view between trials/sessions?             |
 |`referenceTargetColor` |`Color3`               | The color of the "reference" targets spawned between trials                        |
 |`referenceTargetSize`  |m                      | The size of the "reference" targets spawned between trials                         |
-
+|`showPreviewTargetsWithReference` |`bool`      | Show a preview of the trial targets (unhittable) with the reference target. Make these targets hittable once the reference is destroyed |
+|`previewTargetColor`   |`Color3`               | Set the color to draw the preview targets with (before they are active)            |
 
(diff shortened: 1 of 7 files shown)


bboudaoud-nv:
Currently this branch only supports showing preview targets during a "reference" target. There was a generic request to support previewing the next trials targets during the current trial when reference targets are not drawn as well, but this might be a candidate for a follow-up issue/PR unless issue #187 is updated.

jspjutNV:
I tested this with the following session config and the basic functionality appears to work:
```
        { 
            showReferenceTarget = false;
            showPreviewTargetsWithReference = true;
            previewTargetColor = Color3(0.5, 0.8, 0.5);
            description = "training";
            frameRate = 240;
            id = "240Hz";
            settingsVersion = 1;
            trials = (
                { 
                    count = 2;
                    ids = ( "static", "moving", "moving");
                } );
        }
```

PR #201, opened 2020-07-02
Remove the old sample configs

jspjutNV:
The sample configs hadn't been updated in a long time and were out of date, this pull request was intended to update them to work with the latest builds of FPSci. Now it's just removing the files.

The below was originally the plan but no longer is.

Still some things to be done in this PR.

- [ ] turn off prints between trials/sessions for "real" sessions
- [ ] test that these configs actually function

Some things that may or may not make sense to do

- [ ] consider renaming these configs
- [ ] consider whether these configs should be included with automated builds in a `samples` directory or similar


jspjutNV:
Let's just delete these samples.

PR #202, opened 2020-07-08
Foreground and blocking commands

bboudaoud-nv:
This branch adds support for a `CommandSpec` which allows the experiment designer to specify `foreground` and `blocking` flags for any given command.

This slightly modifies the syntax for specifying command arrays in the experiment config file from:

```
commandsOnSessionEnd = [ "testcommand.exe" ];
```
to
```
commandsOnSessionEnd = [ { command = "testcommand.exe" } ];
```

This restructure allows the `foreground` and `blocking` parameters to be specified along with the `command`. For example, to open a foreground browser with a web page at the end of a session, the designer can specify:

```
commandsOnSessionEnd = [ { command = "cmd /c start [webpage URL]", foreground = true } ];
```

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -469,29 +469,49 @@ These flags control whether various information is written to the output databas
 ## Command Config
 In addition to the programmable behavior above the general config also supports running of arbitrary commands around the FPSci runtime. Note that the "end" commands keep running and there's the potential for orphaned processes if you specify commands that are long running or infinite. The command options include:
 
-| Parameter Name                    | Type              | Description                                                                                  |
-|-----------------------------------|-------------------|----------------------------------------------------------------------------------------------|
-|`commandsOnSessionStart`           |`Array<String>`    | Command(s) to run at the start of a new session. Command(s) quit on session end              |
-|`commandsOnSessionEnd`             |`Array<String>`    | Command(s) to run at the end of a new session. Command(s) not forced to quit                 |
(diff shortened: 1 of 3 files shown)


jspjutNV:
I ran the same test as in #152 but modified it for the new syntax and added blocking to the first prints to ensure that both prints would happen. This appears to function as intended. Here's the updated test session:
```
        {
            description = "Test Commands";
            frameRate = 240;
            id = "commands";
            trials = ( {count = 2; ids = ( "static" );}, {count = 2; ids = ("static", "static") } );
            commandsOnSessionStart = ( 
                {command = "cmd /c echo Session start>> commandLog.txt", blocking = true}, 
                {command = "cmd /c echo Session start second command>> commandLog.txt"} 
            );
            commandsOnSessionEnd = ( 
                {command = "cmd /c echo Session end>> commandLog.txt", blocking = true, foreground = true}, 
                {command = "cmd /c echo Session end second command>> commandLog.txt"} 
            );
            commandsOnTrialStart = ( {command = "cmd /c echo Trial start>> commandLog.txt"} );
            commandsOnTrialEnd = ( {command = "cmd /c echo Trial end>> commandLog.txt"} );
        }
```

jspjutNV:
In testing `foreground = true` I've noticed that the mouse continues to be captured even while interacting with the application (web browser in my test) that's running in the foreground. I tested on the trial end, but if done at session end along side bringing up the menu, then this issue should be mitigated.

PR #203, opened 2020-07-08
Refactor system/latency logger config

### LINKED ISSUE #197: Consider removing latency logger (system) config file
We've already greatly reduced the use of system config to just the latency logger config parameters. @jspjutNV has made a good argument for moving away from it's use entirely in favor of experiment config parameters.

This issue is a place to discuss the idea of removing the latency logger/system config file in favor of our new command interface and possible some additional general config parameters.

bboudaoud-nv:
This branch removes the system config (renamed to `LatencyLoggerConfig` in code) in favor of uniting these parameters with the experiment config file. This has become the de-facto standard anyway as other loggers are now using commands specified in the `experimentconfig.Any` file rather than using the system config as intended.

It also adds support for replace-able substrings in commands, notably `%loggerComPort` and `%loggerSyncComPort` which now come from their affiliated input fields in the experiment config file.

Merging this PR closes #197.

### DIFF
--- a/FPSci.lib.vcxproj
+++ b/FPSci.lib.vcxproj
@@ -178,7 +178,6 @@
     <None Include="data-files\keymap.Any" />
     <None Include="data-files\shader\distort.pix" />
     <None Include="data-files\startupconfig.Any" />
-    <None Include="data-files\systemconfig.Any" />
     <None Include="data-files\userconfig.Any" />
     <None Include="data-files\userstatus.Any" />
   </ItemGroup>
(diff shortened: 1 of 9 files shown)


bboudaoud-nv:
Based on discussion in #197 we've decided not to remove the system config, but rather continue to support it to keep the experiment config file as generic as possible. This PR now handles adding support for command substrings, but does not remove the `SystemConfig` class (renamed back from `LatencyLoggerConfig`) or it's use in per-machine setup.

bboudaoud-nv:
We should wait for #196 to be merged, then merge master into this branch before opening this for review.

PR #213, opened 2020-07-15
Test `packages` path in `vs`

jspjutNV:
This is a fix to get the tests looking for the `packages` directory inside of the `vs` directory where it gets placed after the changes from #210 

This fix is only needed for the tests to work correctly.

This is an updated version of #212 since that pull request wouldn't let us merge it (thanks github).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,9 @@
 .vscode
 .vs
 # Don't check in any compiled results
-x64
-Build
-packages
+vs/Build/x64
+vs/Build
+vs/packages
 # Avoid .pyc files (everywhere)
 *.pyc
 # Don't check-in any of these data-files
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -2,9 +2,9 @@
 .vscode
 .vs
 # Don't check in any compiled results
-x64
-Build
-packages
+vs/Build/x64
```
bboudaoud-nv, line 5: If you exclude all of `vs/Build` is this line necessary?

jspjutNV: I don't know. the `x64` was there, which was excluding `Build/x64` as well as `vs/Build/x64`. I don't think `x64` was ever needed, but it was there for some reason. Any idea what that reason is?

bboudaoud-nv: I think `x64` used to be our build directory (historical from when the solution was a single project), I think we now have a more complicated build directory structure that is excluded by the next line.

jspjutNV:
I pushed a removal of the `x64` directory, but it isn't showing in this pull request even though it's in the branch...

PR #217, opened 2020-07-16
Allow packager generator to work without `log.txt`

jspjutNV:
The previous packager generator required a `log.txt` to operate, but it's useful to allow it to run and only generate a new `README.txt` so this pull request adds support for this mode of operation.


bboudaoud-nv:
Looks like reasonable improvements to me

Issue #135, opened 2020-05-27
Settings window visual bugs when allowReticleChange=true


jspjutNV:
There are two visual bugs related to the visual upgrade of the user settings window that only appear when `allowReticleChange = true;` is set in the experiment config.

1. The window is oversized relative to the resume and quit buttons.
2. The window spawns off center.

bboudaoud-nv:
This may come from within G3D, and is possibly related to issue #93.

bboudaoud-nv:
This no longer seems to be occurring. I believe @jspjutNV made some changes that may have impacted this issue. If we can confirm it is gone, let's log the resolution and close this issue.

jspjutNV:
I just tested and the window is still not centered.

The buttons and window sizing at the bottom are still not correct either. This issue should remain open.

jspjutNV:
Here's an image demonstrating the oversized window. Noticed how the right side extends farther past the buttons than the left. Also, the header is not centered.
![image](https://user-images.githubusercontent.com/26633010/84674225-b03afe00-aef8-11ea-9283-be00da9f29e4.png)

jspjutNV:
To demonstrate the non-centering, see the following two images. The first one is how the app boots up. After closing and reopening the menu, it moves to the centered position seen in the second one.
![image](https://user-images.githubusercontent.com/26633010/84674379-dbbde880-aef8-11ea-8931-d1b526319ad5.png)

![image](https://user-images.githubusercontent.com/26633010/84674449-eb3d3180-aef8-11ea-9514-0d99ad6fb59f.png)

This difference is minor and small enough that people probably won't notice, and since it's visual only, I don't think this is worth fixing unless someone really wants to polish the user experience.

PR #218, opened 2020-07-16
Improve user menu packing and positioning

### LINKED ISSUE #175: User settings menu screen location
The user settings menu always opens up at the center of the screen. In some usage, I feel like it would make more sense for it to open in the same location where it last closed (if the user moved it), but I can see reasons why we might want it to be centered (if it's underneath a question between sessions or similar).

Does anyone feel strongly one way or another about this? Should the user settings menu remember its screen space position between closing and opening? Should it always open in the same place?

bboudaoud-nv:
This branch is an attempt to improve the packing and positioning of the user menu.


jspjutNV:
Merging in the latest master, it appears there are additional regressions on the user menu, particularly related to some of the recent changes. The menu currently looks like this.

image

Notice the large empty space on the right side.

jspjutNV:
I've fixed the previous over-large menu. Now I notice that the reticle change is a bit too large as well.

image

jspjutNV:
Looking at #175 and #135, this pull request addresses neither of those issues. I'm going to remove the close relationship between them.

The settings menu still doesn't show up centered on startup, and it also doesn't have any way to retain it's location between opening and closing it. The bug of the window continuing to be oversized with reticle controls enabled also remains as can be seen by the above screenshot.

That said, this pull request is still a good improvement, so I'm going to merge it, it's just that both reference issues will need to remain open.

PR #219, opened 2020-07-16
Split/reuse updateTargetColor() for preview targets

bboudaoud-nv:
This branch re-uses target materials (rather than creating new ones in `TargetEntity::setColor()` for activating targets). This should remove "activation artifacts" that might be caused by lengthy material allocation.

The concept is that the pre-created materials in the `FPSciApp::m_materials` array can be used to "activate" targets rather than creating a new one at runtime.

`TargetEntity::setColor()` is still used for spawning targets as this has not been reported as an issue in the past/is an easy way to create the preview material from within the `TargetEntity`.

### DIFF
--- a/source/ConfigFiles.h
+++ b/source/ConfigFiles.h
@@ -749,6 +749,7 @@ public:
 
 	ArticulatedModel::Specification modelSpec;							///< Model to use for the weapon (must be specified when renderModel=true)
 
+	/** Returns true if firePeriod == 0 and autoFire == true */
 	bool isLaser() {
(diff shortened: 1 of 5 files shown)


bboudaoud-nv:
This branch now also includes a logging bug fix wherein a `non-task` event was logged to the `Player_Actions` table in the results for every frame for which the mouse was held down when the weapon was not in auto-fire mode. 

This appears to have been a result of an `else` statement being nested one layer too shallow in the `missEvent()` callback within `FPSciApp`.

PR #222, opened 2020-07-17
Extend model scaling range w/ scale factor

bboudaoud-nv:
This branch changes the `TARGET_MODEL_ARRAY_SCALING` macro in code to extend the range of the model scaling at some limited cost to scale granularity.

FPSci implements model sizing by:

1. Sizing the model to 1m with a "default" scale factor
2. Applying an additional logarithmic scale following: 
```scale = (1 + TARGET_MODEL_ARRAY_SCALING) ^ (i - TARGET_MODEL_ARRAY_OFFSET)```
where `i` is the index within the array (in the range `[0, Target array length - 1]`)

The "fix" here is to keep the target model array the same size, but increase the `TARGET_MODEL_ARRAY_SCALING` (slightly) to increase the range of target sizes that can be created from within the application. This allows a wider range of target scales to be represented (at lower granularity).

Specifically the request from @joohwankimNV was to make smaller player-space targets (typically the smallest targets as they tend to be located the closest to the user).

The original values were:
* `TARGET_MODEL_ARRAY_SCALING` = 0.2
* `TARGET_MODEL_ARRAY_OFFSET` = 20
* Target array length = 30

This meant that the minimum representable target size was given by: `1.2 ^ -20` and corresponds to a target about 0.025m in size. By changing the `TARGET_MODEL_ARRAY_SCALING` to `0.4` we now get a minimum target size of `1.4 ^ -20` or about 0.001m (should be smaller than a pixel at the default view distance on a 1080p monitor).  This represents a minimum target size reduction of 25x. Similarly the max target size increases from about 5m to about 20m. 

The corresponding effect is a drop in the "granularity" of target size, which is dependent on the target size in question. For our most common player-space target sizes (0.01-0.05m) there are still 5 valid target sizes. This should be tested to make sure the granularity is sufficient. If not we can always increase the target array size instead of just changing scaling.

### DIFF
--- a/data-files/test/experimentconfig.Any
+++ b/data-files/test/experimentconfig.Any
@@ -15,6 +15,7 @@
     playerGravity = Vector3(0,0,0);
 
 	showMenuOnStartup = false;
+    showMenuBetweenSessions = false;
 
     weapon = {
         id = "default"; 
@@ -32,6 +33,17 @@
                 {
                     ids = ( "keepalive", "front", "right" );
                     count = 10;
+                }
+            );
+        },
+        {
+            id = "sizes";
+            trials = (
+                {
+                    ids = ( "small", "medium", "large", "toosmall", "toolarge");
+                    //ids = ( "small");
+                    //ids = ( "small", "small", "small", "small", "small");
+                    count = 5;
                 }
             );
         }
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by jspjutNV** [comment]
I've got some target-size-related test cases that I plan to push to this branch shortly. We can iterate on the design, but I want to point out that with the 5 tests I currently have, the proposed change has these effects:

```
abstract-fps\tests\FPSciTests.cpp(522): error: The difference between smallsize and 0.05408785f is 0.0056875869631767273, which exceeds e, where
smallsize evaluates to 0.048400264233350754,
0.05408785f evaluates to 0.054087851196527481, and
e evaluates to 9.9999997473787516e-06.
abstract-fps\tests\FPSciTests.cpp(527): error: The difference between mediumsize and 0.09346383f is 0.0014006868004798889, which exceeds e, where
mediumsize evaluates to 0.094864517450332642,
0.09346383f evaluates to 0.093463830649852753, and
e evaluates to 9.9999997473787516e-06.
abstract-fps\tests\FPSciTests.cpp(532): error: The difference between largesize and 2.07360029f is 0.11360037326812744, which exceeds e, where
largesize evaluates to 1.9599999189376831,
2.07360029f evaluates to 2.0736002922058105, and
e evaluates to 9.9999997473787516e-06.
abstract-fps\tests\FPSciTests.cpp(537): error: The difference between toosmallsize and 0.02608403f is 0.024888833169825375, which exceeds e, where
toosmallsize evaluates to 0.0011951968772336841,
0.02608403f evaluates to 0.026084030047059059, and
e evaluates to 9.9999997473787516e-06.
abstract-fps\tests\FPSciTests.cpp(542): error: The difference between toolargesize and 5.15978241f is 15.501260757446289, which exceeds e, where
toolargesize evaluates to 20.661043167114258,
5.15978241f evaluates to 5.1597824096679688, and
e evaluates to 9.9999997473787516e-06.
```

Notice that the differences between the `small`, `medium` and `large` test cases appear to mostly be differences in rounding from the change and could be made to pass in the test by increasing the epsilon (`e`).

The `toosmall` and `toolarge` have much larger error (relative to their baseline value), which is more expected with the expanded range caused by this change.

With some cleverness, it should be possible to get the `small`, `medium` and `large` to not change much. `toosmall` and `toolarge` will probably want to change the test since we're expanding the range of valid sizes.

bboudaoud-nv:
@jspjutNV suggested we consider increasing the `FPSciApp::modelScaleCount` to 40 and updating the `TARGET_MODEL_ARRAY_OFFSET` macro in `TargetEntity.h` accordingly (say to 30). This produces a similar effect by increasing the total stored targets by ~33%. `TARGET_MODEL_ARRAY_SCALING` is no longer changed.

If we are okay with this change it keeps compatibility with previous target sizes in the app, at the cost of extra memory (not much of a problem for our default target model).

jspjutNV:
It may be worth adding along side this a feature that prints to the log when a target with the maximum or minimum size has been created, just to let the user know what is happening.

jspjutNV:
The new set of changes allows targets ranging from 0.1 mm up to 8 m in size. The tests now check for values that match the config for all except the values outside of the valid range which check for correctly clamped values.

This should be ready to merge now.

PR #223, opened 2020-07-20
Lower audio latency (requires G3D r7012 or newer)

jspjutNV:
This pull request taps into the newly exposed G3D ability to control the DSP buffer size, which affects the audio latency. At the moment, it's unclear what cost these changes have, so I'm opening this pull request up with the intent to test it.

The smaller buffer sizes proposed to reduce latency, but at the cost of increase CPU utilization, which I haven't yet characterized. We need to examine how much that cost is to decide whether it's worth paying, or if we want a less (or more) aggressive setting.

### DIFF
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -19,9 +19,11 @@ int main(int argc, const char* argv[]) {
 		G3DSpecification spec;
 		spec.audio = FPSciApp::startupConfig.audioEnable;
 		
-		// lower audio latency
+		// lower audio latency to 256 / 48000 * 1000 * (3 - 1.5) = 8 ms
+		// Based on Average latency (ms) = bufferlength / 48kHz * 1000ms/s * (numbuffers - 1.5)
+		// See more here: https://github.com/NVlabs/abstract-fps/pull/223
 		spec.audioBufferLength = 256;
-		spec.audioNumBuffers = 1;
+		spec.audioNumBuffers = 3;
 
 		initGLG3D(spec);
 	}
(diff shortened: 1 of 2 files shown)


bboudaoud-nv:
Diving a bit deeper into [docs](https://documentation.help/FMOD-Ex/FMOD_System_SetDSPBufferSize.html) it looks like the average expected latency is set by:

```
Average latency (ms) = blocksize / 48kHz * 1000 * (numblocks - 1.5)
```

So for the default values average latency should be about 53.33ms.

jspjutNV:
Before merging this, someone needs to test out the CPU runtime penalty of lowering the audio latency. 

I'm guessing 128/2 will be better than the current 256/1.

jspjutNV:
I used the math @bboudaoud-nv posted above, and decided to target ~10 ms.

Given a buffer size of 256, 3 and 4 buffers give 8 ms and 13.33 ms. I decided to go with 3 in the latest push. This should be ready for review/merge.

jspjutNV:
Oh, and I tested the CPU impact and didn't see any change on my Core i7 6700K.

bboudaoud-nv:
Agreed, I don't see much CPU impact vs the current `master` implementation w/ default values. I did a quick "click-to-bang measurement here on my home system to validate the end-to-end latency reduction here.

My setup didn't use a gaming mouse/240Hz setup so exact latencies are not to be trusted, but through a few trials it seems click-to-bang audio latency is reduced from about 130ms to 65ms as a result of this update. More precise measurement should be used to verify ground-truth click-to-bang on various systems.

**Original click-to-bang latency**
image

**Click-to-bang latency with this update**
image

PR #225, opened 2020-07-22
Allow developers to reload their configs from disk while running

### LINKED ISSUE #127: Reload Config Files While Running
@jspjutNV pointed out that it would be nice to be able to reload configs on the fly while running the application (particularly in developer mode) similar to how shaders/textures can be reloaded by pressing the F5 key in many/most G3D applications.

This would enable an experiment designer to keep the application up while editing config files and simply "reload" to test new behavior. Seems like a solid enhancement/time saver for those actively working on experiment design.

I'd propose we hard-code this to a key (available in developer mode only) to avoid possible issues with key map reloading interacting with changing the reload key itself.

Creating an issue to host discussion of this feature/it's usefulness prior to starting any implementation.

bboudaoud-nv:
This branch adds support for a new `reloadConfigs` key within the key map. When `developerMode=true` in the startup config, this allows developers to reload configs from within the application without restarting. By default this key is set to `F5` to coincide with the generic reload offered by G3D.

Merging this PR closes #127

### DIFF
--- a/docs/developermode.md
+++ b/docs/developermode.md
@@ -16,6 +16,7 @@ Features include:
 * Dynamic brightness (exposure) control
 * Waypoint-based [target path creation](./patheditor.md)
 * Initialization of player position and view direction
+* Dynamic config reload in app via the `reloadConfigs` mapped key
 
 ## Enabling Developer Mode
 As mentioned above, all that needs to be done to enabled developer mode is modifying the `developerMode` field in [`startupconfig.Any`](../data-files/startupconfig.Any) to `True`, then running the application to enter developer mode.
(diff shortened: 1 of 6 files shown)


bboudaoud-nv:
Currently this PR handles the added feature of reloading configs as well as a "fix" for G3D function keys still keeping their pre-bound values when `developerMode = false` in the startup config.

These fixes include setting the `GApp::Settings::useDeveloperTools` value to the startup config's `developMode` flag and also "masking off" the `F8` key stroke used for capturing a cube map (in non-developer mode, this feature is still available in developer mode).

PR #229, opened 2020-07-28
Logging improvements for analysis

### LINKED ISSUE #226: User config turn scale logging ignores session setting
Currently we log the (per user) `cmp360` and `turnScale` values to the `Users` table in the database at the start/end of every session, but the per-session turn scale is not included in this value. 

Should we consider logging this as well/providing an "effective" cm/360° sensitivity (composite of all turn scales w/ sensitivity) in this/another table?

### LINKED ISSUE #227: Target logging improvements
Current the output database's `Targets` table only logs parameters (typically randomized within ranges) used to spawn a target, rather than the actual value of the parameters used for these targets. This is particularly relevant to targets where `destSpace = player`.

While the current approach allows target info to be logged to the targets table just once per session, it creates 2 issues:
1. The exact size/spawn eccentricities/speed(s) of a target over a trial are unknown without using the `Target_Trajectory` table and spatial math to attempt to reconstruct this information
2. The set of targets spawned within a given trial are not explicitly provided without using the `Target_Trajectory` table

Should we consider logging to the `Targets` table once per trial with more detailed information for the randomized values we've chosen for each target spawn? Should we consider a target naming convention that uniquifies down to the individual targets spawned within a trial?

None of this need break compatibility with previous analysis if added correctly. The suggested improvements effectively amount to a quality of life improvement in data analysis efforts.

bboudaoud-nv:
There are a number of things we could improve around our current results database logging approach, including, but not limited to:

* Improving the per-trial target information  to the `Targets` table #227 
    * Add actual/randomized size of the target
    * Add actual/randomized spawn eccentricity for the target
* Including the per-experiment/session `turnScale` parameter in the `Users` table logging #226
    * Add/delineate the user and session `turnScale` in the log
    * Add a composite X/Y sensitivity to the log (computed as the product of turn scales and `cmp360` setting)
    * Make the individual targets identifiable to a single entry in the `Trials` table (likely by timestamp)
* Improve/create the documentation of our output logs so we have a place to direct users for reading/interacting with the logs

This PR is a place to discuss and review all of the changes mentioned above.

Merging this PR closes #226 and closes #227


bboudaoud-nv:
In addition to the improvements listed and made above, some other topics for discussion on logger improvement include:

- [ ] Removing the `refresh_rate` and `added_frame_lag` fields from the `Targets` table in favor of using `paramsToLog`/the `Sessions` table for this purpose
- [ ] Providing angular data on target size/position (may be difficult with view projection artifacts)
- [ ] Establishing a fully unique trial naming scheme so that timestamps are not needed to affiliate trials to (some) logs
- [ ] Adding user columns to more tables so that logs can be merged between simultaneous trials (assumed different users)
- [ ] Adding the `reticleChangeTimeS` to the user log columns
- [ ] Adding other relevant/non-parameter defined info to the sessions table (i.e. total trials/blocks)

jspjutNV:
I'm not sure about each specific proposal here, but I believe the databases should be made to be fully specified and referenced using foreign keys that you can always identify which user run which sessions and experiments, and which specific trial data belongs to a particular user's execution of the given experiment/session. 

With sufficient specificity, merging of databases should be relatively straightforward.

bboudaoud-nv:
I would suggest we create another issue/PR to handle "merge safety" for DBs (I've explicitly tried to avoid it in this PR). 

My goal here is primarily "quality of life" style logging improvements as opposed to allowing multi-db merge across various sessions for a single user/multiple users who might have taken part in sessions simultaneously.

This is a good topic for discussion, but it could live in another request (worked on serially after this one?).

bboudaoud-nv:
The `refresh_rate` and `added_frame_lag` parameters have been dropped from the `Targets` table and added to the default value for `sessParamsToLog` such that they are logged to the `Sessions` table by default.

Note that in this implementation if you specify the `sessParamsToLog` array yourself in the config you will need to add `"frameRate"` and `"frameDelay"` to the list if you want them to be included in the `Sessions` table (i.e. by specifying `sessParamsToLog = []` you can eliminate these parameters from the `Sessions` table).

bboudaoud-nv:
In addition we've agreed to: 

- [ ] Move the target position logging to world space (not relative to spawn location)
- [ ] Move player heading to world space (not relative to spawn heading)

This will make all target/player positions consistent in world space and remove the need to know the initial player heading in order to extract the relationship between the player position/view direction and target within that view.

docs/general_config.md
```diff
@@ -188,7 +188,7 @@ As part of the general configuration parameters several controls over reporting
 "logFrameInfo": true,
 "logPlayerActions": true,
 "logTrialResponse": true,
-"sessParamsToLog" : [],
+"sessParamsToLog" : ["frameRate", "frameDelay"],
```
jspjutNV, line 191: I think `sessParamsToLog` should be renamed to `sessionParametersToLog` to match the style of the rest of things about sessions and parameters. This is the only use of Parameters that I can find, so we could shorten to `Params`, but I prefer the couple extra characters for clarity.

source/Session.cpp
```diff
@@ -434,17 +435,12 @@ void Session::accumulateTrajectories()
 {
 	if (notNull(m_logger) && m_config->logger.logTargetTrajectories) {
 		for (shared_ptr<TargetEntity> target : m_targetArray) {
-			if (!target->isLogged()) continue;
-			// recording target trajectories
-			Point3 targetAbsolutePosition = target->frame().translation;
-			Point3 initialSpawnPos = m_camera->frame().translation;
-			Point3 targetPosition = targetAbsolutePosition - initialSpawnPos;
-					   
+			if (!target->isLogged()) continue;
```
jspjutNV, line 438: This change should be validated to function the same. Is there a currently existing test to make sure this isn't a regression?

bboudaoud-nv: This particular line isn't a change, but I assume the reference is to the camera relative vs absolute coordinate change below? If so this is _not_ intended to function the same. Per previous discussion targets are now logged in world-space as opposed to relative to the camera frame

bboudaoud-nv:
Consider adding a `Target Types` table that provides high-level target info independent of spawned targets. Discuss whether a `Trial Types` table also makes sense.

source/Session.cpp
jspjutNV: Okay, I realize I was looking at this in the wrong context. Looks good.

docs/general_config.md
bboudaoud-nv: This should now be changed.

PR #234, opened 2020-08-19
Fixes for Pointer-based Interaction in the Dialog/Menu System

### LINKED ISSUE #230: Mouse cursor hides on question prompts
I've had it reported by users, and have noticed myself an intermittent bug where sometimes on transitioning to a question response phase of a session the mouse cursor doesn't show. I can sometimes work around this bug by opening the menu.

I'll attempt to figure out how to reproduce this bug and post the steps soon.

bboudaoud-nv:
This branch attempts to correct bugs around pointer vs camera control interaction of the mouse in various experiment configurations.

Merging this PR closes #230.

### DIFF
--- a/source/FPSciApp.cpp
+++ b/source/FPSciApp.cpp
@@ -348,7 +348,7 @@ void FPSciApp::presentQuestion(Question question) {
 	}
 
 	moveToCenter(dialog);
-	this->addWidget(dialog);
+	addWidget(dialog);
 	setDirectMode(false);
 }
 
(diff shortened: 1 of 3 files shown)


bboudaoud-nv:
I've marked this ready for review, since the change in question is useful/ready, but we should verify there aren't additional edge cases where interaction between windows hides the mouse when it shouldn't be.

jspjutNV (approved):
I say we merge now, and we can always open a new fix if we notice more problems in the future.

PR #239, opened 2020-09-21
Moved to deg/mm mouse sensitivity

### LINKED ISSUE #238: Use degrees/mm to set mouse sensitivity
Given that cmp360 is confusing to most users, we should switch the displayed value and slider in the user settings to use degrees/mm instead of cmp360. 

If we're going to use deg/mm, we should probably change the internal representation to deg/mm. Let's add a display only box to the UI to show the cmp360 to make it easier for people to map between online calculators. Might be worth showing the equation there as well so they can do the conversion if needed.

I think we should include a feature that allows loading cmp360 if that is specified in the user settings file, and converting that into deg/mm.

bboudaoud-nv:
This PR tracks merging in °/mm (as opposed to historic cm/360°) mouse sensitivity. 

This is a good idea for 2 reasons:

1. °/mm mouse sensitivity measures _increase_ at greater sensitivity (as you'd expect)
2. °/mm and cm/360° are tightly related (`°/mm = 36 / (cm/360°)`)

Merging this request closes #238


bboudaoud-nv:
Currently implementation removes the user config's `cmp360` setting in favor of a `mouseDegPerMillimeter` setting that's been added. 

cm/360° sensitivity is still available through a button press in the user menu.

jspjutNV (approved):
This generally looks good now. I'd be okay merging it now, unless @bboudaoud-nv wants to change the logger to log deg/mm instead of, or in addition to cmp360. It seems like we're at the point that we basically want to be seeing deg/mm everywhere, and the logs haven't been updated yet.

jspjutNV:
I've updated this to log both parameters. I proposed we schedule removing the cmp360 logging for 6 months from now and consider it deprecated. If I'm counting right, that would be an April 2021 release at the latest.

PR #245, opened 2020-10-02
Add new users

bboudaoud-nv:
This branch adds support for adding new users from within the application. As it a result it also adds:

- A user configuration flag to require unique user IDs (`requireUnique` in the user config)
- A user menu field/button for entering/adding new user IDs and a field providing feedback on empty/duplicate user IDs
- Support for automatically switching to the added user

Once new users are created the user can be directed to setup their mouse sensitivity and/or reticle (depending on what is allowed for the experiment) from the normal user menu.

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -460,6 +460,7 @@ These flags control the display of the in-game user menu:
 |`showMenuLogo`                     |`bool`     |Show a logo at the top of the menu (currently `materials/FPSciLogo.png`) |
 |`showExperimentSettings`           |`bool`     |Show the options to select user/session                                |
 |`showUserSettings`                 |`bool`     |Show the per-user customization (sensitivity, reticle, etc) options    |
+|`allowUserAdd`                     |`bool`     |Allow users to add new users to the experiment                         |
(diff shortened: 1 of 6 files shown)


bboudaoud-nv:
Currently the DPI and reticle index are borrowed from the last user in the users list. We could implement additional configuration parameters for the high level user table that provide defaults here.

jspjutNV:
I notice that the shift key still fires at the dummy target if it's present. This makes typing capital letters in a user name a little annoying, but not show stopping. I'll probably merge this if it's the only problem, and open a separate issue to fix the problem.

jspjutNV:
Actually, I fixed the dummy shot problem, continuing my review.

jspjutNV:
I've noticed a regression where this crashes when a selected user has completed all sessions when `selectedSession()` is called in `GuiElements.h`. This appears to be related to the change adding `.text()` to the return value.

I suspect that the logic in the function may have changed incorrectly since `m_ddCurrSessIdx` appears to be -1 in this case, and the `>=` comparison means the empty case will never be caught.

jspjutNV (approved):
After the handful of small fixes, this looks ready to merge.

PR #248, opened 2020-10-09
Add (optional) bullet spread

bboudaoud-nv:
This branch adds support for bullet spread via the `fireSpread` configuration parameter in the weapon configuration. By default this spread value is set to 0, maintaining the purely deterministic shot behavior previously used in FPSci.

Currently the fire "spread" is shot-to-shot and uniformly distributed/sized in azimuth/elevation with no effects of things like the time since the last shot was made. If we'd like we can extend this to include:

- Impact of time since the last shot
- Asymmetric azimuth and elevation spread
- Support for multiple projectiles in a single shot

This branch does not currently support changing the damage of the weapon based on shot spread (i.e. distance from central aim direction) in any way.


bboudaoud-nv:
A recent commit adds support for a `fireSpreadShape` parameter that allows distributional shaping of the fire spread. Default value is `"uniform"` but support for `"gaussian"` is also available.

jspjutNV (approved):

PR #249, opened 2020-10-13
Weapon Improvement

bboudaoud-nv:
This branch makes the `Weapon` class more self-managing, and also adds support for a (visual only) weapon kick via the new `kickAngle` and `kickDuration` parameters.

### DIFF
--- a/source/FPSciApp.cpp
+++ b/source/FPSciApp.cpp
@@ -260,21 +260,42 @@ void FPSciApp::updateControls(bool firstSession) {
 	if (startupConfig.waypointEditorMode) { waypointManager->updateControls(); }
 
 	// Update the player controls
-	if(notNull(m_playerControls)) removeWidget(m_playerControls);
+	bool visible = false;
+	Rect2D rect;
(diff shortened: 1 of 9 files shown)


bboudaoud-nv:
Merging this PR should now close #259

source/FPSciApp.cpp
```diff
@@ -260,21 +260,42 @@ void FPSciApp::updateControls(bool firstSession) {
 	if (startupConfig.waypointEditorMode) { waypointManager->updateControls(); }
 
 	// Update the player controls
-	if(notNull(m_playerControls)) removeWidget(m_playerControls);
+	bool visible = false;
```
jspjutNV, line 263: Are these changes through about line 300 the ones that cause menus to remember their previous locations?

bboudaoud-nv: Yes, the local variables (`visible` and `rect`) store the visibility state and window rectangle (size/position) respectively while we remove/re-create the controls widgets

PR #253, opened 2020-10-30
Fix(es) for memory leak(s) in new session logic

bboudaoud-nv:
This branch attempts to reduce a memory leak that appears to be caused by creating new sessions within the application. It also includes general code cleanup that improves the consistency and readability of FPSciApp code.


source/FPSciApp.cpp
```diff
@@ -495,6 +497,7 @@ void FPSciApp::updateSession(const String& id) {
 	}
 
 	// Create a series of colored materials to choose from for target health
+	m_materials.clear();
```
jspjutNV, line 500: @bboudaoud-nv says this is the primary memory leak fix.

PR #255, opened 2020-10-30
Add allowSessionChange control flag

bboudaoud-nv:
This branch adds an `allowSessionChange` control flag to disable users from updating the session from the user menu.


jspjutNV (approved):
Looks good.

PR #257, opened 2020-11-16
Support unified (per user) experiment results file

bboudaoud-nv:
Support a unified user/experiment results database for cleaner output.

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -517,10 +517,10 @@ These flags control whether various information is written to the output databas
 "logPlayerActions" = true,
 "logTrialResponse" = true,
 "logUsers" = true,
-"logToSingleDb" = false,
+"logToSingleDb" = true,
 ```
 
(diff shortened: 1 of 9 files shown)


bboudaoud-nv:
This branch doesn't (currently) have support for experiments that use different `sessParamsToLog` between sessions. This could potentially be supported using some clever logic in the future.

jspjutNV:
I think the default should be to log to a single file by default (once we get this working as intended).

We also should think through the "single log file" design in case someone is working on multiple revisions of the same experiment. We may want something to uniquify the file based on an particular experiment. Maybe we could append a hash of the `experimentconfig.Any`.

jspjutNV:
We also should think through the "single log file" design in case someone is working on multiple revisions of the same experiment. We may want something to uniquify the file based on an particular experiment. Maybe we could append a hash of the `experimentconfig.Any`.

bboudaoud-nv:
A hash (of a reformatted experiment config, that ignores whitespace/invalid parameters) has been added to the filename here. If @joohwankimNV agrees that a unified experiment db then we can make this change and merge this PR.

jspjutNV (approved):

PR #258, opened 2020-11-19
Question Improvements

bboudaoud-nv:
Add support for key-bindings for question responses and fullscreen questions.


jspjutNV (approved):
Looks good

PR #260, opened 2020-12-03
Reorganization of SQL and Database

### LINKED ISSUE #372: Issue with 'trialsComplete' column in 'Sessions' table
The count in trialsComplete column in Sessions table is lower than the true trial counts.

I have N target types, M repetitions/block, and L blocks. The correct count I expect is N x M x L, but I only see M x L.

jspjutNV:
This pull request will include the following:

- [x] Reorganize SQL queries in C++ to make them easier to debug
- [x] Add an Experiment table for future situations where multiple experiments may need to merge into a single database
- [x] Use a consistent "True" vs "true" capitalization in database entries for binary values.

Benefits of these changes include:

1. Easier iteration on database schema updates
2. Ability to merge results for an experiment across multiple users and PCs
3. Potential to reuse the same or similar schema for storage in other types of databases

Merging this branch also closes #372


bboudaoud-nv:
I've tried to make a number of small quality of life improvements in this code including better use of `INTEGER` types and use of the new `BOOLEAN` type as well.

Before closing this PR we should harmonize on either using or not using `sqlHelpers.cpp` and either remove it or revert `Logger::createSessionsTable()` to use `createTableInDB()` accordingly.

bboudaoud-nv:
I've moved back to using `createTableInDB()` for `Logger::createSessionsTable()`. I think it is a decent idea to have our database access run through an intermediate layer, but agree with @jspjutNV we should better document this. In addition I've tried to avoid passing large `Array`s to our sqlHelpers methods

bboudaoud-nv:
I've added an early version of an `Experiments` table to our results file. As discussed w/ @jspjutNV this table includes the experiment description, time it was logged, and its hash, as well as the full text of the experiment config (making the experiment recoverable).

Currently there should only ever be one entry in this table b/c the experiment config hash is included in the results filename when using unified logging, and each session is given its own results file otherwise. In the future we may want to consider checking whether the experiment hash we compute is in this table before we add it (and potentially add experiments to the table other than when we first create the database).

jspjutNV:
This PR is going to cause errors in SQL output for people who had run an experiment on a previous version of FPSci, then run the same experiment on this version. We'll want to document this in the release notes. An example of the errors that will show up in the `log.txt` is the following:

```
Error in UPDATE statement (): no such column: trials_complete
```

The fix is to delete or rename the results file (such as `TestExperiment_anon_9d971c91.db`) and only then run FPSci.

This is because we changed a bunch of the column names and the old database file will have incompatible column names (and numbers of columns) as a result.

PR #262, opened 2020-12-04
Experiment selection

### LINKED ISSUE #261: Startup config to allow specifying a list of experiments
It would be a nice feature to allow specifying multiple different experiment configs in the startup config. I propose a syntax like the following:

```
    experimentList = (
        {
            name = "my cool experiment";
            experimentConfigFilename    = "experiments/experimentconfig.Any"; 
            userConfigFilename          = "experiments/userconfig.Any"; 
            userStatusFilename          = "experiments/userstatus.Any"; 
        },
        {
            name = "another good experiment";
            experimentConfigFilename    = "experiments/experimentconfig2.Any"; 
            userConfigFilename          = "experiments/userconfig.Any"; 
            userStatusFilename          = "experiments/userstatus2.Any"; 
        },
    );
```

In addition to being able to specify this list of experiments, we should also have a GUI dropdown menu to select among them while running in developer mode. 

Note that a `userstatus.Any` file creates the explicit correspondence between a `userconfig.Any` and an `experimentconfig.Any`, so experiment configs and user configs may be reused in this list.

bboudaoud-nv:
This branch adds support for selecting the experiment from an `experimentList` when in developer mode.

The approach is a bit complicated, but should make things easier for users. The only config changes are in the startup config file, which now supports 2 new fields (and drops support for "direct" configuration file specification):

- A `defaultExperiment` which sets up the files to use as defaults for experiment(s) if none are provided, and specifies the only experiment if no `experimentList` is provided
- An `experimentList` which provides a list of (named) experiments as deviations from the `defaultExperiment` provided above

The `defaultExperiment` and `experimentList` entries contain the same file fields that used to exist at the top level of the startup config, namely:

- `experimentConfigFilename`
- `userConfigFilename`
- `userStatusFilename`
- `keymapConfigFilename`
- `systemConfigFilename`
- `resultsDirPath`

In addition to these fields, entries in the `experimentList` also require a `name` field to be provided to make them unique.

When run in developer mode FPSci allows the developer to select from an additional user menu drop-down list of experiments. Otherwise the app (currently) always runs the first experiment in the `experimentList` (or the `defaultExperiment` if no/an empty `experimentList` is provided).

Merging this PR closes #261

### DIFF
--- a/docs/developermode.md
+++ b/docs/developermode.md
@@ -5,6 +5,7 @@ During development of `FirstPersonScience` we created a number of utilities for 
 When running in developer mode a number of the otherwise static features of the tool are available for dynamic toggle via an expanded Heads Up Display menu accessible during the normally available pause (Tab) menu.
 
 Features include:
+* Switching between multiple experiments (as specified in the `experimentList` within (startup config)[startupConfigReadme.md])
 * Dynamic control of bullet rendering
 * Dynamic control of HUD rendering
 * Dynamic control of on-screen FPS monitoring
--- a/docs/startupConfigReadme.md
+++ b/docs/startupConfigReadme.md
@@ -24,6 +24,7 @@ The following fields are specified on a per-experiment basis:
 * `keymapConfigFilename` sets the path and file name for a [keymap config file](./keymap.md).
 * `systemConfigFilename` sets the path and file name for a [system config file](systemConfigReadme.md) for controlling latency logger hardware.
 * `resultsDirPath` sets the path to the results directory. If this directory does not exists the application will create it at runtime.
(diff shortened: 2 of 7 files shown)


jspjutNV:
This is already looking pretty good, but I think it's got some more improvements that we should make. I'm going to schedule this for the February release.

bboudaoud-nv:
Changes are welcome, there are still some occasional bugs occurring around what seems to be `GApp` widget management (specifically event processing) when we re-initialize the experiment. We might want to consider an alternate approach to cleaning up state before we initialize the experiment.

jspjutNV (approved):
Looks good

PR #265, opened 2020-12-16
Remove duplicate insert of playerCamera into scene

### LINKED ISSUE #264: Two entities with same name "defaultCamera"
When running in the `debug` configuration an initial (ignorable) exception dialog is created with the message in the title of this issue. This may be due to G3D creating a `defaultCamera` and us also using this name in our default `scene.Any` file.

bboudaoud-nv:
This branch fixes a bug from `FPSciApp::onAfterLoadScene()` wherein the `playerCamera` was being inserted into the scene after the scene is loaded (which already inserts the camera). This exception only showed up when running the application in the `Debug` configuration, and was ignore-able, but caused confusion for some developers.

Merging this PR closes #264.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,9 +22,17 @@ data-files/test/userconfig.Any
 "scripts/event logger/software/Logs"
 results/
 # This avoids checking-in experiment-specific files
-data-files/sound
-data-files/scene
-data-files/model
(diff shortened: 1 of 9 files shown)


jspjutNV (approved):
Looks good to me

PR #268, opened 2021-01-19
Fix target sounds not playing and add new default sounds

### LINKED ISSUE #267: Target hit sound not playing
A common feature in aim trainers is to have a hit confirmation sound that is different than the fire sound. Right now FPSci allows a miss sound to be played, but doesn't allow a hit sound to be specified. 

I believe it would be best to specify this sound per target, which would allow us to have different sounds for "body" or "headshots" potentially.

While we're doing this, we should double check that the miss sound is working correctly. It didn't appear to work in my quick test.

jspjutNV:
This PR closes #267 

This includes the following:
* Playing target hit sounds regardless of the weapon `firePeriod`.
* More precision in the rounding of `firePeriod` and `damage` (1 ms and 1% of target health/s).
* New audio samples in `sound/fpsci_*.wav`
  * In addition to the below, `sound/fpsci_destroy_300ms.wav` is available for longer destroy sounds
  * In addition to the below, `sound/fpsci_noise_50ms.wav` is useful for continuous-fire weapons
* Switches the default sounds
  * Hit targets use `sound/fpsci_ding_100ms.wav` 
  * Destroyed targets use `sound/fpsci_destroy_150ms.wav` 
  * Weapon fire uses `sound/fpsci_fire_100ms.wav` 
  * Scene hit use `sound/fpsci_miss_100ms.wav`
* All sound volumes now default to 100%


bboudaoud-nv:
When targets are destroyed (i.e. damaged below 0 health w/o respawn) the hit audio isn't played. It seems like the audio play call is made, but the object is destroyed before it can complete playing. We should add a delayed destruction mechanic to prevent this from happening.

jspjutNV:
> When targets are destroyed (i.e. damaged below 0 health w/o respawn) the hit audio isn't played. It seems like the audio play call is made, but the object is destroyed before it can complete playing. We should add a delayed destruction mechanic to prevent this from happening.

It seems like the destroyed sound also isn't playing. If possible, we should add a fix to both of these sounds not playing issues to this pull request.

jspjutNV:
Commit [34bf347](https://github.com/NVlabs/abstract-fps/pull/268/commits/34bf347df9ac84acf56c6b2080b1633cda8a2f26) fixes the destroyed and hit sounds not playing when the target is destroyed at the same time.

bboudaoud-nv:
The `soundTable` approach is a good one to avoid issues like the one mentioned in #267. 

I will approve this for now, but for the future we should consider:
1. Supporting loading sounds ahead of trials (i.e. in `updateSession()`) rather than at target spawn time
2. Storing the `soundTable` with the `Session` (rather than `FPSciApp`) to avoid memory bloat
3. Clearing the `soundTable` between sessions if we choose to keep it in `FPSciApp`

PR #273, opened 2021-02-09
Moves weapon firing to onSimulation

jspjutNV:
In #269 the new tests don't yet pass, and it made sense to do a rather large restructuring of the code around weapon firing. This pull request moves the weapon firing into `onSimulation` instead of `onUserInput` so that it can be decoupled.

In its current state, this partly works, but it still doesn't pass all the tests, and there's likely some additional issues still present. I'm creating another pull request into the main PR in case this path ends up being unproductive and we decide to do something else.

### DIFF
--- a/source/Weapon.cpp
+++ b/source/Weapon.cpp
@@ -201,7 +201,7 @@ void Weapon::drawDecal(const Point3& point, const Vector3& normal, bool hit) {
 
 void Weapon::clearDecals() {
 	while (m_currentMissDecals.size() > 0) {
-		m_scene->remove(m_currentMissDecals.pop());
+		m_currentMissDecals.pop();
 	}
 	if (notNull(m_hitDecal)) {
 		m_scene->remove(m_hitDecal);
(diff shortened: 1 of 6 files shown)


source/Weapon.cpp
```diff
@@ -201,7 +201,7 @@ void Weapon::drawDecal(const Point3& point, const Vector3& normal, bool hit) {
 
 void Weapon::clearDecals() {
 	while (m_currentMissDecals.size() > 0) {
-		m_scene->remove(m_currentMissDecals.pop());
+		m_currentMissDecals.pop();
```
jspjutNV, line 204: This fixes a long standing bug where there'd be a crash occasionally when switching sessions or experiments.

jspjutNV:
This PR made a rather large structural change to the way weapon firing is handled to support `firePeriod` values near the frame period with autofire enabled. This change moved to a model where the mouse state change is all that is handled in `FPSciApp::onUserInput()` and the weapon firing dynamics in `FPSciApp::onSimulation()` consider that state before choosing how many bullets to fire, and how much damage each bullet should apply should it hit a target. (Most of this happened in #273 )

This change is in anticipation of potential future needs where the simulation, input and rendering may need more flexibility. A future PR will probably move from using real, wall-clock time for the weapon dynamics to using the simulation time that is kept by `FPSciApp::oneFrame()`.

PR #278, opened 2021-02-19
Startup config loading

### LINKED ISSUE #277: FPSci crashes if no `startupconfig.Any` is present
You can reproduce this by downloading [v21.02.01](https://github.com/NVlabs/abstract-fps/releases/tag/v21.02.01) and double clicking `FirstPersonScience.exe`. On first run, it will create the `startupconfig.Any` and crash. The second run after the file is created works as expected.

bboudaoud-nv:
This branch introduces a `StartupConfig::load()` static method that handles checking for a valid startup config, creating one if not available, then returning the resulting object. This solves the issue of an exception at startup when running without a valid `startupconfig.Any` file in the `data-files` directory (i.e. when a new `startupconfig.Any` is created).

Merging this PR closes #277.

### DIFF
--- a/source/ConfigFiles.h
+++ b/source/ConfigFiles.h
@@ -105,6 +105,14 @@ public:
 	
 	bool	audioEnable = true;									///< Audio on/off
 
+	static StartupConfig load(const String& filename) {
+		StartupConfig config;
+		if (!FileSystem::exists(filename)) {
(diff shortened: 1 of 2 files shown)


jspjutNV:
Looks good. Works in my test.

jspjutNV (approved):

PR #286, opened 2021-02-25
Null check for scene hit sound

bboudaoud-nv:
Fixes an issue where an empty string value for `sceneHitSound` in the experiment/session config resulted in a null pointer exception at runtime.


bboudaoud-nv:
If we'd like we can do an additional update of documentation in this branch to formally add support for empty strings as a way to disable various audio around FPSci.

jspjutNV (approved):

bboudaoud-nv:
Documentation should now be updated to officially recommend empty string to disable audio.

PR #290, opened 2021-03-03
Experiment config validation

### LINKED ISSUE #289: Specifying a target id in a session that isn't in the target list crashes
Steps to reproduce:

1. put a target ID that isn't in the target list into a session trial
2. Start up that session

We should detect this problem earlier and provide a helpful error message.

bboudaoud-nv:
This branch adds support for an `ExperimentConfig::validate()` routine that is called (optionally) on `ExperimentConfig::load()`. Specifically this validate routine checks that the set of target ids used in session `TrialCount` objects match the global list of targets specified in the experiment config `targets` array.

Merging this PR closes #289.


bboudaoud-nv:
Leaving this as a draft in case there are other things we want to add to the experiment config validation routine. If not, then feel free to mark ready/review.

jspjutNV:
I moved the validation up a level, and I added an initial validation loop of all startup config experiments in the list. This lets some experiments fail to load gracefully and only allows validated experiments to remain on the list. This should be ready now.

jspjutNV (approved):

PR #292, opened 2021-03-04
Default user to userconfig.Any

### LINKED ISSUE #291: `defaultUser` is removed from user config when written
Steps to reproduce:

1. User config needs a filled in `defaultUser`
2. Experiment config needs `allowUserAdd = True;`
3. Start FPSci
4. Add a new user

The `defaultUser` field will have been removed from the user config. 

As far as I can tell the default user continues to exist in memory while FPSci is still running, so additional users added before quitting will continue to copy the default user settings. However, on subsequent start up and runs, the previous default user settings will have been forgotten.

bboudaoud-nv:
This branch adds support for serializing the `defaultUser` of the user config back out when a user is updated/created.

Merging this PR closes #291.

### DIFF
--- a/source/UserConfig.cpp
+++ b/source/UserConfig.cpp
@@ -110,8 +110,9 @@ UserTable UserTable::load(const String& filename) {
 
 Any UserTable::toAny(const bool forceAll) const {
 	Any a(Any::TABLE);
-	a["settingsVersion"] = 1;						///< Create a version 1 file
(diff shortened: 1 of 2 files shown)


jspjutNV (approved):
This looks like it should work. I only worried about the change to the default setting of `forceAll`, but it looks like we were ignoring that argument before anyway in this case.

PR #293, opened 2021-03-11
Clearing miss decals on new sessions/trials

bboudaoud-nv:
This branch adds:

- A quick fix for the `Weapon::clearDecals()` call that removes decals from the scene
- A new `clearTrialMissDecals` weapon config flag (defaults to `true`) that allows experiment designers to choose whether miss decals are removed at the start of a new trial
- A new `missDecalTimeoutS` weapon config parameter (default to `-1` for no timeout) that allows experiment designers to set a (simulation time-based) timeout for miss decals

### DIFF
--- a/source/Weapon.cpp
+++ b/source/Weapon.cpp
@@ -309,7 +309,8 @@ void Weapon::drawDecal(const Point3& point, const Vector3& normal, bool hit) {
 
 void Weapon::clearDecals() {
 	while (m_currentMissDecals.size() > 0) {
-		m_currentMissDecals.pop();
+		shared_ptr<VisibleEntity> decal = m_currentMissDecals.pop();
+		m_scene->remove(decal);
 	}
 	if (notNull(m_hitDecal)) {
 		m_scene->remove(m_hitDecal);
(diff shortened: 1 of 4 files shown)


source/Weapon.cpp
```diff
@@ -300,18 +317,23 @@ void Weapon::drawDecal(const Point3& point, const Vector3& normal, bool hit) {
...
 		m_hitDecalTimeRemainingS = m_config->hitDecalDurationS;
 	}
 }
 
-void Weapon::clearDecals() {
-	while (m_currentMissDecals.size() > 0) {
-		m_currentMissDecals.pop();
+void Weapon::clearDecals(bool clearHitDecal) {
+	while (m_currentMissDecals.size() > 0) {				// Remove and clear miss decals
+		m_scene->remove(m_currentMissDecals.pop());
```
jspjutNV, line 332: This line is crashing for me on reload (F5). I assume it's because the scene isn't the same one that the `m_currentMissDecals` assumed it was.

jspjutNV (changes requested):
I believe `hitDecalDuration` and `missDecalTimeoutS` are accomplishing the same thing, just on different decals. Let's unify naming by switching `hitDecalDuration` to `hitDecalTimeoutS`. As far as I know, nobody has used that feature, but we'll still document it in the release notes.

source/Weapon.cpp
bboudaoud-nv: Up to now I haven't been able to reproduce this crash. Any other details on repro are welcomed.

jspjutNV: Since I want to include this in the UL experiment, let's go ahead an merge with this bug pending. It really only impacts developers, so I'm sure we'll hit it and fix it in the future.

PR #295, opened 2021-03-16
Logger null check in Session::processResponse()

### LINKED ISSUE #294: Crash when logEnable = False
When `logEnable` is `False` the app crashes following completing any trial.

bboudaoud-nv:
Quick null check addition to prevent crash in `Session::processResponse()`

Merging this PR closes #294.


bboudaoud-nv:
Leaving this as a draft to check for additional places where logger might be null

jspjutNV (approved):

PR #299, opened 2021-03-24
Added symmetric eccentricity support to player space target spawn positions

ArjunMadhusudan:
This branch adds more support for defining player space targets' spawn areas. The symmetric eccentricity (`symmetricEcc`) flag can be set to false (true by default) in a target description, which allows for setting signed spawn boxes based on absolute values provided. (as opposed to the existing bi-axially symmetrical spawn points being used right now)


jspjutNV:
Hey @ArjunMadhusudan , could you add a brief description of what this pull request is intended to accomplish beyond just the title?

Thanks!

source/Logger.cpp
```diff
@@ -370,6 +370,7 @@ void FPSciLogger::logTargetTypes(const Array<shared_ptr<TargetConfig>>& targets)
 			"'" + config->destSpace + "'",
 			String(std::to_string(config->size[0])),
 			String(std::to_string(config->size[1])),
+			String(std::to_string(config->symmetricEcc)),
```
jspjutNV, line 373: I believe this adds the new parameter to each row added to the table, but I don't think the column labels have been updated, so this is likely to cause a problem. We'll need to make sure the column list gets updated before merging this PR.

docs/experimentConfigReadme.md
```diff
@@ -96,6 +96,7 @@ The following configuration only applies to player-bound parametric targets.
 
 * `speed` is a vector indictating the minimum ([0]) and maximum ([1]) speeds in angular velocity (in deg/s)
 * `distance` is the distance to this target (in meters)
+* `symmetricEcc` is a boolean flag indicating whether the target spawn eccentricities are to be considered in all 4 directions or not
```
bboudaoud-nv, line 99: I suggest improving this description text, possibly including an example. Something along the lines of:

>When `True` eccentricities are assumed symmetric about 0° (the central view direction) and should always be >0. When `False` the eccentricities can be signed, with positive eccentricities indicate values to the right (azim) and up (elev) of the central view direction.

docs/experimentConfigReadme.md
```diff
@@ -127,6 +128,7 @@ targets = [
         "visualSize" : [0.5, 0.5],              // 0.5m size
         "respawnCount" : 0,                     // Don't respawn
         "speed": [1.0, 3.0],                    // 1-3m/s speed
+        "symmetricEcc" : false;                 // Target will only spawn on one quadrant according to eccH/V signs
```
bboudaoud-nv, line 131: As a convention we try to document the default values in these examples, so I would make this `True`

source/Logger.cpp
bboudaoud-nv, line 373: You need to add this column to the table creation at line 75 above in this file for this to work

bboudaoud-nv (changes requested):
Overall good job with this change, it was a great place to get started on FPSci development. With some minor documentation improve and a minor logger bug fix this should be merge-able.

jspjutNV:
@ArjunMadhusudan and I think it would be good to have independent horizontal and vertical control with 2 parameters instead of one.

ArjunMadhusudan:
Added support for axial symmetry as per discussion with @jspjutNV.

bboudaoud-nv (approved):

bboudaoud-nv:
Assuming you've tested logging to make sure the columns now work in the output log and run the `FPSci.test` project (within the solution) to make sure no tests fail here that don't also fail on `master` I'm fine w/ merging these changes.

source/Logger.cpp
```diff
@@ -370,6 +372,8 @@ void FPSciLogger::logTargetTypes(const Array<shared_ptr<TargetConfig>>& targets)
 			"'" + config->destSpace + "'",
 			String(std::to_string(config->size[0])),
 			String(std::to_string(config->size[1])),
+			String(std::to_string(config->symmetricEccH)),
```
bboudaoud-nv, line 375: This results in a string value of "0" or "1" being written to the file instead of "true" or "false". You can see examples of how we setup these strings in `updateSessionEntry()` or just above this in `logTargetTypes()` (for the `jumpEnabled` value). I'd suggest moving to this style of string reporting here.

Having said that we are inconsistent in our use of capitalization in these "true" and "false" fields, so maybe we should create an issue to harmonize this for the entire logger class.

bboudaoud-nv:
Other than the one issue mentioned above this branch is ready for merge, all tests that pass on `master` pass here.

source/Logger.cpp
jspjutNV: I agree it's a problem. I'll add a check box to the big database restructuring PR #260  and we can roll the fix in there.

ArjunMadhusudan: Fixed!

PR #300, opened 2021-03-26
Cache session shaders manually to work around LAUNCH_SHADER assumption

jspjutNV:
The G3D `LAUNCH_SHADER` macro has a `static` variable inside, assuming the same shader will be provided every time that macro is used. This PR switches to the `LAUNCH_SHADER_PTR` macro that allows us to control shader loading and caching explicitly.

Fixes a bug where sessions with different custom shaders would always use whichever shader was loaded first.


bboudaoud-nv:
I made some minor changes to use `FPSciApp::updateSession()` to (re)load the shader before each session. This prevents against doing the load in an `onGraphics()` routine.

PR #310, opened 2021-07-27
Add `samples` directory for sample experiments

### LINKED ISSUE #307: Provide collection of sample experiments
A fresh FPSci download would be a better experience if it came with a set of sample experiments that a developer could start from. This is an issue to track progress toward improving these sample experiments.

jspjutNV:
This PR adds a `samples` directory which is scanned on startup to add provided sample experiments to the `experimentList`. This is primarily a developer-oriented feature, and won't be noticed when `developerMode = false`. 

Closes #307


jspjutNV:
I've left this as a draft until I add at least 1-2 samples. But this scan and include functionality should work already.

jspjutNV:
This should be mostly ready. I did notice a minor issue that if the set of files changes in the samples directory and the user presses F5 while FPSci is running, then the old list of samples will remain. This is likely an existing issue with startupconfig changes, and unlikely to matter since users probably won't be adding new samples.

data-files/samples/targets.Experiment.Any
```diff
@@ -0,0 +1,129 @@
...
+                { 
+                    count = 6; 
+                    ids = ( "jumping");
+                }, 
+            ); 
+            
+        }, 
+    ); 
+    
+    settingsVersion = 1;
```
bboudaoud-nv, line 97: I usually specify `settingsVersion` at the top of the file (or more realistically don't specify it at all)

data-files/samples/sample.User.Any
```diff
@@ -0,0 +1,11 @@
+{ 
+    settingsVersion = 1; 
+    users = ( 
+        { 
+            cmp360 = 20;
```
bboudaoud-nv, line 5: Would we rather specify `mouseDegPerMillimeter` here? Both work, but I believe `mouseDegPerMillimeter` overrides `cmp360` if specified.

data-files/samples/targets.Experiment.Any
```diff
@@ -0,0 +1,129 @@
...
+            jumpEnabled = true; 
+            jumpPeriod = ( 0.5, 1 ); 
+            jumpSpeed = (8,8);
+            accelGravity = (20, 20);
+            motionChangePeriod = ( 0.8, 1.5 ); 
+            speed = ( 10, 10 ); 
+            visualSize = ( 0.05, 0.05 ); 
+        } ); 
+    
+    weapon = {
```
bboudaoud-nv, line 125: This is a duplicate definition w/ the weapon at the top of this file. Unless this is deliberate we should probably remove this. If it is deliberate we should add a comment explaining why its defined twice.

data-files/samples/targets.Experiment.Any
```diff
@@ -0,0 +1,129 @@
...
+            ); 
+            
+        }, 
+    ); 
+    
+    settingsVersion = 1; 
+    targets = ( 
+        { 
+            destSpace = "player"; 
+            id = "static";
```
bboudaoud-nv, line 101: I usually put the ID at the top of the target definition to help make things clear. Fine to reorder to show you can if that's the goal here.

data-files/samples/targets.Experiment.Any
```diff
@@ -0,0 +1,129 @@
...
+    settingsVersion = 1; 
+    targets = ( 
+        { 
+            destSpace = "player"; 
+            id = "static"; 
+            speed = ( 0, 0 ); 
+            visualSize = ( 0.05, 0.05 ); 
+        }, 
+        { 
+            axisLocked = ( false, false, true );
```
bboudaoud-nv, line 106: I could be wrong but I believe axis locks only apply to world space targets...

data-files/samples/targets.Status.Any
```diff
@@ -0,0 +1,12 @@
+{ 
+    currentUser = "Sample User"; 
+    sessions = ( "1target", "2targets", "5targets", "moving", "jumping", "mixedtypes"); 
+    settingsVersion = 1; 
+    users = ( 
+        { 
+            completedSessions = ( );
```
bboudaoud-nv, line 7: As a general comment, once this `completedSession` list is full this sample will just show "All Sessions Completed". I assume this is fine, but right now we don't really have a good mechanism for clearing out this array...

data-files/samples/targets.Status.Any
```diff
@@ -0,0 +1,12 @@
+{ 
+    currentUser = "Sample User"; 
+    sessions = ( "1target", "2targets", "5targets", "moving", "jumping", "mixedtypes"); 
+    settingsVersion = 1; 
+    users = ( 
+        { 
+            completedSessions = ( ); 
+            id = "Sample User"; 
+            sessions = ( "1target", "2targets", "5targets", "moving", "jumping", "mixedtypes");
```
bboudaoud-nv, line 9: Specifying this and the `sessions` array above in the same order is a bit odd, perhaps include a comment to describe why. Generally speaking I usually just spec on or the other.

data-files/samples/weapons.Experiment.Any
```diff
@@ -0,0 +1,151 @@
+{ 
+    description = "manytargets";
```
bboudaoud-nv, line 2: Description could be something about different weapons?

data-files/samples/weapons.Experiment.Any
```diff
@@ -0,0 +1,151 @@
...
+    frameRate = 120; 
+    showHUD = true;
+    renderWeaponStatus = true;
+    cooldownMode = "ring";
+    cooldownColor = Color4(0.7,1.0,0.9,0.75);
+    sessions = ( 
+        { 
+            description = "1 shot"; 
+            id = "1shot"; 
+            settingsVersion = 1;
```
bboudaoud-nv, line 12: Again settings version really doesn't _need_ to be specified here, but if it is I usually do `id, description, settings version` at the top of the session...

data-files/samples/weapons.Experiment.Any
```diff
@@ -0,0 +1,151 @@
...
+            trials = ( 
+                { 
+                    count = 20; 
+                    ids = ( "static", "static", "moving"); 
+                } ); 
+            
+            weapon = { 
+                id = "2hit"; 
+                firePeriod = 0.3;
+                damagePerSecond = 2;
```
bboudaoud-nv, line 40: Could add a comment describing how this adds up to a 2-hit kill since the math isn't "exact" (i.e. perfect would be 1.6666...) but could be 1.7 with a comment

data-files/samples/weapons.Experiment.Any
```diff
@@ -0,0 +1,151 @@
...
+                    ids = ( "static", "static", "moving"); 
+                } ); 
+            
+            weapon = { 
+                id = "show"; 
+                firePeriod = 0.3;
+                damagePerSecond = 1;
+                autoFire = false;
+                renderModel = True;
+                modelSpec = ArticulatedModel::Specification{			        // Default model
```
bboudaoud-nv, line 121: Doing the full model spec here is descriptive, but may be overwhelming. IIRC technically you just need `renderModel = True` to get something drawn.

data-files/samples/weapons.Experiment.Any
```diff
@@ -0,0 +1,151 @@
+{ 
+    description = "manytargets"; 
+    frameRate = 120; 
+    showHUD = true;
+    renderWeaponStatus = true;
+    cooldownMode = "ring";
+    cooldownColor = Color4(0.7,1.0,0.9,0.75);
```
bboudaoud-nv, line 7: You could also do something like change cooldown color per session here as a demo of cascaded config if you'd like

data-files/samples/weapons.Experiment.Any
```diff
@@ -0,0 +1,151 @@
...
+                        transformGeometry(all(), Matrix4::scale(1.2,0.7,0.4));
+                    };
+                    scale = 0.25;
+                },
+            }; 
+            
+        } 
+    ); 
+    
+    settingsVersion = 1;
```
bboudaoud-nv, line 134: Again a bit far down for `settingsVersion`, but probably not a deal breaker

data-files/samples/weapons.Experiment.Any
```diff
@@ -0,0 +1,151 @@
...
+    settingsVersion = 1; 
+    targets = ( 
+        { 
+            destSpace = "player"; 
+            id = "static"; 
+            speed = ( 0, 0 ); 
+            visualSize = ( 0.05, 0.05 ); 
+        }, 
+        { 
+            axisLocked = ( false, false, true );
```
bboudaoud-nv, line 143: Same as previous, I believe `axisLocked` doesn't have meaning for player-space targets

data-files/samples/weapons.Experiment.Any
bboudaoud-nv: I'd suggest removing the cooldown indicator altogether for the autofire weapon sessions below

bboudaoud-nv (changes requested):
Overall this is a positive set of examples for new developers to reference, but I think there are a few changes we should consider before merging this PR.

data-files/samples/sample.User.Any
jspjutNV: Agreed

data-files/samples/targets.Status.Any
jspjutNV: I agree this would be a practical problem for trying out the samples, but I don't really have a good solution for this. They can always revert the file 🤷‍♂️

data-files/samples/targets.Status.Any
jspjutNV: The one above is in case an experiment ends up allowing adding new users. I'm not sure it's all that important, and the functionality is documented in the user status documentation.

data-files/samples/weapons.Experiment.Any
jspjutNV: While true, I think this is a nicer sample since it gives people the hooks they'd need to replace the model if they want to. In any case, this is a supplement to the docs, and hopefully not a replacement.

data-files/samples/weapons.Experiment.Any
jspjutNV: I intentionally left it in for the autofire, but I took it out for the fastfire one. I think 200ms fire period is about the max speed at which a cooldown indicator makes sense, though it's definitely experiment specific.

data-files/samples/weapons.Experiment.Any
jspjutNV: I'm trying to mostly have these configs "speak for themselves" and leave the comments in the documentation. That said, I don't think we explicitly call out the need to worry about rounding on weapon damage. I'll check.

docs/weaponConfigReadme.md
```diff
@@ -39,6 +39,8 @@ This file provides information about the weapon to be used in the experiment. De
     "fireSpreadShape": "uniform";   // Uniform shape of fire spread distribution by default
 ```
 
+Note that the `firePeriod` and `damagePerSecond` interact to create the amount of damage done per shot to the target. A target has 1.0 health, and will be eliminated once the damage is greater than or equal to that value. When deciding on these values, it's useful to set the `damagePerSecond` slightly higher than the pure multiple to avoid floating point error accumulating to require an additional shot. As an example, if you want 5 shots to kill a target over 1 second, you may be tempted to set the values to `firePeriod = 0.2` and `damagePerSecond = 1.0`, but it's possible that it will take 6 shots and 1.2 seconds to eliminate the target. By setting `damagePerSecond = 1.1`, it will ensure that the target will be killed in 5 shots over 1.0 seconds.
```
bboudaoud-nv, line 42: Based on discussion w/ @jspjutNV I think we agree these shot counts are right but times are off by 1 shot. 6 shots ==> 1s and 5 shots ==> 0.8s (or slightly more) as there are N-1 cooldowns in N shots.

bboudaoud-nv (approved):

PR #311, opened 2021-07-27
Add support for random in range `pretrialDuration`

### LINKED ISSUE #309: Random range of `pretrialDuration`
We'd like to allow the `pretrialDuration` to be specified as a range so that the target will spawn after a random delay within that range. A single value should continue to function as it always has, the target spawns after that fixed duration.

bboudaoud-nv:
This branch adds support for random-in-range `pretrialDuration`. This does not break previous configs by allowing experiment designers to specify _either_ a scalar value or a 2-array of values for randomization in range. This array is expected to be formatted `[min value, max value]`.

Exceptions are raised when the specified array is undersize (of length 0 or 1), and warnings are printed to the log when oversized (of length >2).

Merging this PR closes #309.

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -57,7 +57,8 @@ The following settings allow the user to control various timings/durations aroun
 | Parameter Name                |Units              | Description                                                        |
 |-------------------------------|-------------------|--------------------------------------------------------------------|
 |`clickToStart`                 |`bool`             |Require the user to click at the start of the session to spawn the first reference target  |
-|`pretrialDuration`             |s (or `Array<`s`>`)  |The time before the start of each trial (or a randomized range)     |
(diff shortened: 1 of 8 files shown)


jspjutNV:
[Nickerson and Burnham in 1969 published "Response Times with Nonaging Foreperiods"](https://psycnet.apa.org/record/1969-07662-001) ([pdf](https://www.researchgate.net/profile/Raymond-Nickerson/publication/23611777_Response_times_with_nonaging_foreperiods/links/57476c4b08ae707fe21e3ea1/Response-times-with-nonaging-foreperiods.pdf)) which makes the case that a uniform random distribution is less good for this purpose than a truncated exponential distribution. 

I'm referencing this source in case we manage to update this PR with a truncated exponential distribution.

jspjutNV (approved):

Issue #313, opened 2021-08-02
Additional methods for keeping score and informing users


jspjutNV:
Right now, score is primarily kept within a trial, and for a few things (like remaining time) across a session. It would be useful to be able to keep an experiment-designer-defined score both running across trials within a session as well as across sessions within an experiment. 

This defined score would then be reportable to the user feedback messages as well as to the database. If a shared leaderboard is ever implemented, then this score would also be reported to that leaderboard, though that functionality is purely theoretical at this point.

Open topics for discussion before beginning implementation:
* How to express what values contribute to "score"
  * What values can contribute toward score (hits, misses, time taking, time remaining, etc.)
* How the score should be reported to the user
* Any potential issues with logging scores to the database? Does the score expression need to be logged? Do the sources consumed to create the score get logged?

bboudaoud-nv:
This is an interesting and useful feature idea, but the bounds of what could be done here are quite broad. Options I would consider include:

- User specified, string-based "score expressions" that might include math done between one or more "base metrics" we provide. These could become arbitrarily complex and require a nearly complete math library to be defined. For example `score = "max(1/taskTimeS, 100) * shotsHit / (shotsHit + shotsMissed)`. Alternatively this could be simplified to only allow particular, predefined "base metrics" for score
- Working from @jspjutNV's initial list I'd say a good set of base metrics (for trials or sessions) would include:
    - hits
    - misses
    - targets destroyed
    - damage done
    - total shots
    - accuracy (as ratio or percentage)
    - task time
    - time remaining
    - time spent firing (as a total or, more likely,  a ratio)
    - trial successes (session only)
    - trial failures (session only)
- I believe score should be reported to the user in 1 of 2 core ways:
    - Via the in-game banner (where different field visibility should probably be controllable, i.e. score but not time)
    - Through formatted feedback messages
- I agree that w/ more configurability logging score to the database isn't a bad idea (you could support logging the score expression through the `sessParamsToLog` array); however, you should certainly be able to reproduce any score metric from the results written to the database (this might be a good test of what our db is missing "easy access" to).

PR #314, opened 2021-08-02
Improve Banner

bboudaoud-nv:
This branch updates the banner to improve configuration and view. Notably it adds:

- Support for `bannerShowTime`, `bannerShowProgress`, and `bannerShowScore` configuration fields to allow elements to be individually enabled
  - `bannerShowTime` now supports `"elapsed"` and `"remaining"` modes (as well as `"none"` for disable) 
- Support for formatting large scores in easier to read ways (use of `k`, `M`, `B` suffixes)
- Support for drawing `---` as time when this value is greater than 10k seconds
- Only showing the time field when valid (when in the `trialTask` state)

These changes are related to the changes requested in #313.


jspjutNV:
Am I correct in my reading of this that it doesn't do anything to support different types of scores as requested by #313 and instead is just about displaying the current scoring mechanism in a few different ways? Just checking so we know to be sure to keep #313 open if we merge this PR.

The mention of a "banner" also reminds me that we should probably do a better job describing what the banner is in the documentation. A labeled screenshot might be nice.

The `bannerShowTime` variable makes me thing that it's setting when, and for how long to show the banner, but it looks like the variable actually controls how to format the countdown (or up) timer. I'd propose the following variable name changes:

* `bannerEnable` - currently `showBanner`
* `bannerTimerMode` - currently `bannerShowTime`
* `bannerProgressEnable` - currently `bannerShowProgress`
* `bannerScoreEnable` - currently `bannerShowScore`

I recognize this would break some existing configs using `showBanner` so we might want to add a warning message if `showBanner` is found in a config.

bboudaoud-nv:
Correct, this PR isn't linked to/doesn't close #313 as it does not propose a solution for that issue, just cleans up general banner behavior. I believe we should create a separate PR to deal with #313 once we decide on a more formal design for scoring.

I agree that adding a screenshot to document the banner is a good idea. This screen shot should probably include labels of the time, progress, and score fields.

We use the `show[X]` convention is fairly widely in FPSci config parameter names. The `showHUD`, `showAmmo`, `showPlayerHealth`, `showReferenceTarget`, `showCursor` etc parameters is what `showBanner`/`bannerShow[X]` naming was based upon, so if do choose to change this we should likely change everything. I believe the only place we currently use `[x]Enable` is the `logEnable` field (which makes sense since it is not a visual element).

I do think it is a good idea to move all of our config fields to names that start with `banner` if possible, grouping them logically if serialized out. I also agree that `bannerTimerMode` is an improvement over `bannerShowTime` and we should make this change. Otherwise I don't have strong preferences on naming for these fields.

jspjutNV:
Cool, your reasoning makes  sense. Let's just switch to `bannerTimerMode` and keep the other names how they were.

jspjutNV:
I just tried to run this PR on my experiment config that works with `master` and other branches, and this one crashes on startup with an access violation in line 181 of `Logger.cpp`. I can't explain what is going wrong, or why these specific changes would cause this error (on a cursory read of them they shouldn't), but this is going to need more investigation.

bboudaoud-nv:
@jspjutNV can you provide the exception message and your config file to review?

This exception seems odd since nothing in this branch should impact the logger at all...

jspjutNV:
Repeating my comment from #319 since it's probably the same bug I'm hitting.

I believe I've gotten it to build by changing line 182 in `Logger.cpp` to make the first `"'"` literal into a `G3D::String` with the following.

```c++
		String("'") + sessConfig->id + "'",
```

Since there's a bunch more of the same implicit conversion to G3D strings to get the G3D `+` operators to work, it seems like we should probably be explicitly making those strings into `G3D::String` types. I cannot explain why this only breaks on a couple branches, some of which were working for me before, but doesn't break on the master branch.

jspjutNV (approved):
The only unexpected thing here was the change in `getScore()` to `100*`, but @bboudaoud-nv explained that it was merging two different `*10` before it was displayed.

PR #316, opened 2021-08-10
Update Any output to be JSON compatible

### LINKED ISSUE #315: JSON Incompatible Any Outputs
When calling `Any::save(filename)` the file output is assumed to not be JSON compatible unless the provided `filename` ends with a `.json` extension.

Incompatibilities include:
- Parameter names are not surrounded w/ quotes
- `=` is used instead of `:`
- Lines end with `;` instead of `,`
- Arrays use `( ... )` instead of `[ ... ]`

G3D can support a "quick fix" here by adding an (optional) `json` argument to `Any::save()` (as is done for `Any::serialize()`) or we can write our own method. If @jspjutNV can get this added to G3D its likely easier to do the former and add a `true` for this new parameter to make Any's serialized out from FPSci JSON parse-able.

If we do this perhaps we should consider harmonizing all of our example docs on the JSON standard formatting...

bboudaoud-nv:
This branch updates all FPSci Any outputs to be JSON compatible using the new 2nd argument of the `Any::save()` method.

Make sure to update and build G3D with `svn up` from the G3D install directory then `./buildg3d lib` from the `G3D10` directory within the install directory prior to building these changes.

Merging this PR closes #315

### DIFF
--- a/docs/startupConfigReadme.md
+++ b/docs/startupConfigReadme.md
@@ -14,7 +14,7 @@ The following fields are valid for a startupconfig.Any file:
 * `defaultExperiment` controls the default set of files to use for the experiment, these values are inherited by any unspecified values from experiments in the `experimentList` (see below)
 * `experimentList` optionally specifies a list of experiments that can be selected from in developer mode, if none is provided a single experiment that matches the `defaultExperiment` specification is used
 * `audioEnable` turns on or off audio
-* `jsonOutput` writes all outputs to JSON-compatible .Any files
+* `jsonAnyOutput` writes all config outputs as JSON-format .Any files
 
 ## Experiment Specification
 The following fields are specified on a per-experiment basis:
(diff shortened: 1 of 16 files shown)


jspjutNV:
I'm thinking we probably want a different method for tracking upstream changes to G3D, but for the moment, let's include which specific revision of G3D this is intended to use.

Based on a quick skim of [G3D commits](https://sourceforge.net/p/g3d/code/HEAD/tree/), it looks like [r7167](https://sourceforge.net/p/g3d/code/7167/) and newer should be compatible with this PR.

jspjutNV:
I tested this with G3D revision r7167 and it passed the tests after compiling successfully. 

However, I personally prefer the C++ style `.Any` format, so I'm not sure why we should force JSON compatible `.Any` output universally. Is there a reason to turn this on all the time other than just making the data files explicitly JSON?

As an example, I typically set my editor to use C++ syntax highlighting for `.Any` and it works quite well, plus I can copy-paste data from C++ into `.Any` and vice versa.

bboudaoud-nv:
This branch doesn't stop users from authoring Any files using any syntax they would like, though it will overwrite changes, particularly those in user status, with a JSON compatible version.

I suppose we could support only serializing the user config/status file, though the only time the startup/experiment config are saved is when there doesn't exist a file on disk. We could also make this contingent on a (startup?) config flag, though there might be some places that gets complicated. Likely worth discussing...

The primary target was making (written to) Any's directly compatible w/ JSON so they can be easily parsed/generated and output by other language's (like Python) directly. There is no parser I'm aware of for Any's "full" syntax in anything but C++. This can make data import/analysis painful. IMO moving towards more "pure" JSON isn't much of a downside for external collaborators as the full compatible Any syntax isn't well enough documented and syntax support outside of G3D is non-existent.

PR #319, opened 2021-09-20
Support for programmable variable frame rate

jspjutNV:
This PR adds support for specifying frame rate variability through one of the following approaches. This PR adds two render parameters, `frameTimeArray` and `randomFrameTime`, both of which are ignored if `frameTimeArray` is empty, which is the default.

* Specify a frame time sequence to follow by placing that sequence in `frameTimeArray` and setting `randomFrameTime = false`.
* Specify a frame time distribution to choose from randomly by placing the list of frame times in `frameTimeArray` and setting `randomFrameTime = true`.

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -145,7 +145,7 @@ allSessionsCompleteFeedback: "All Sessions Complete!",
 |`horizontalFieldOfView`    |°      | The (horizontal) field of view for the user's display, to get the vertical FoV multiply this by `1 / your display's aspect ratio` (9/16 for common FHD, or 1920x1080)|
 |`frameDelay`               |frames | An (integer) count of frames to delay to control latency           |
 |`frameRate`                |fps/Hz | The (target) frame rate of the display (constant for a given session) for more info see the [Frame Rate Modes section](#Frame-Rate-Modes) below.|
-|`frameTimeArray`           |`Array<float>`| An array of frame times to use instead of `frameRate` if populated, otherwise ignored. |
+|`frameTimeArray`           |`Array<float>`| An array of frame times (in seconds) to use instead of `frameRate` if populated, otherwise ignored. |
 |`randomFrameTime`          |`bool` | Whether to selected items from `frameTimeArray` sequentially, or as a uniform random choice. Ignored if `frameTimeArray` is empty. |
 |`resolution2D`             |`Array<int>`| The resolution to render 2D content at (defaults to window resolution)       |
 |`resolution3D`             |`Array<int>`| The resolution to render 3D content at (defaults to window resolution)       |
@@ -169,7 +169,7 @@ For more information on G3D `Sampler` options refer to [this reference page](htt
 "horizontalFieldOfView":  103.0,            // Field of view (horizontal) for the user in degrees
 "frameDelay" : 3,                           // Frame delay (in frames)
 "frameRate" : 60,                           // Frame/update rate (in Hz)
-"frameTimeArray" : [],                      // Array of frame times to use instead of `frameRate` if not empty
+"frameTimeArray" : [],                      // Array of frame times (in seconds) to use instead of `frameRate` if not empty
 "randomFrameTime" : false,                  // Choose items from `frameTimeArray` in order
 
 "resolution2D": [0,0],                      // Use native resolution for 2D by default
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -146,7 +146,8 @@ allSessionsCompleteFeedback: "All Sessions Complete!",
 |`frameDelay`               |frames | An (integer) count of frames to delay to control latency           |
 |`frameRate`                |fps/Hz | The (target) frame rate of the display (constant for a given session) for more info see the [Frame Rate Modes section](#Frame-Rate-Modes) below.|
 |`frameTimeArray`           |`Array<float>`| An array of frame times (in seconds) to use instead of `frameRate` if populated, otherwise ignored. |
-|`randomFrameTime`          |`bool` | Whether to selected items from `frameTimeArray` sequentially, or as a uniform random choice. Ignored if `frameTimeArray` is empty. |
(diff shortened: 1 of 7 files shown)


jspjutNV:
This PR touches some of the inner workings of our frame rate control, and in particular, I haven't yet tested how it interacts with frame delay settings. It's possible that we'll need to do some fixing to make them work together, though my cursory look appeared to work fine.

jspjutNV:
I'm also not committed to the variable naming I chose, nor the function placement necessarily. Given that these parameters belong to the `RenderConfig` but the `FPSciApp` is requesting the target frame time from the `Session` the functionality could really go anywhere in that stack. Happy to hear other opinions on variable naming.

jspjutNV:
As a test case, I used the following settings with `randomFrameTime` switched true and false. This gives you half a second of 10 fps, then half a second of 30 fps, which are noticeable enough to see the change. With `randomFrameTime = true` it gives the 10 Hz stutter randomly, which is noticeably different than the half-second on/off.

```
            frameTimeArray = ( 0.1, 0.1, 0.1, 0.1, 0.1, 0.033, 0.033, 0.034, 0.033, 0.033, 0.034, 0.033, 0.033, 0.034, 0.033, 0.033, 0.034, 0.033, 0.033, 0.034, );
            randomFrameTime = false;
```

docs/general_config.md
```diff
@@ -145,6 +145,8 @@ allSessionsCompleteFeedback: "All Sessions Complete!",
 |`horizontalFieldOfView`    |°      | The (horizontal) field of view for the user's display, to get the vertical FoV multiply this by `1 / your display's aspect ratio` (9/16 for common FHD, or 1920x1080)|
 |`frameDelay`               |frames | An (integer) count of frames to delay to control latency           |
 |`frameRate`                |fps/Hz | The (target) frame rate of the display (constant for a given session) for more info see the [Frame Rate Modes section](#Frame-Rate-Modes) below.|
+|`frameTimeArray`           |`Array<float>`| An array of frame times to use instead of `frameRate` if populated, otherwise ignored. |
```
bboudaoud-nv, line 148: Would be good to include the units for these frame times here. Could do it in the units field as `Array<float>` s or `Array<`s`>`, or just specify it in the description. I assume these times are specified in seconds?

source/Session.cpp
```diff
@@ -121,6 +121,29 @@ bool Session::hasNextCondition() const{
...
+			return 1.0f / float(m_app->window()->settings().refreshRate);
+		}
+	}
+	else {
+		if (m_config->render.randomFrameTime) {
+			return m_config->render.frameTimeArray.randomElement();
+		}
+		else {
+			static uint targetIdx = 0;
+			return m_config->render.frameTimeArray[targetIdx++ % m_config->render.frameTimeArray.size()];
```
bboudaoud-nv, line 141: I think the indexing here might introduce a minor bug/benefit from some simplification.

Since `targetIdx` is a static uint that is never reset it will eventually wrap around the max uint value. This will produce a discontinuity in the frame times. Though extremely uncommon, I think this can be solved w/ the following code:

```
static uint targetIdx = 0;
targetIdx += 1;
targetIdx %= m_config->render.frameTimeArray.size();
return m_config->render.frameTimeArray[targetIdx];
```

This prevents the overflow "discontinuity" event by  capping `targetIdx` at the frame time array size. I guess technically we should also document that there is a 2^32-1 limit on the length of the frame array, but even at 1kHz frame rate this is still something like 50 days of consecutive values... so likely not an issue.

source/Session.cpp
```diff
@@ -121,6 +121,29 @@ bool Session::hasNextCondition() const{
...
+		}
+	}
+	return 0.0;
```
bboudaoud-nv, line 144: Seems like this would never be hit, I suppose this value would indicate "don't wait at all" correct?

bboudaoud-nv (changes requested):
Seems like a good set of changes overall. I like this approach for specifying a set of frame times and allowing support for random sampling, lets us do a sort of "inverse CDF" type approach to model almost any distribution!

I suppose we should write some validation scrips that look at the `sdt` arrays in the db file to make sure things are "working" as expected. Otherwise w/ one minor change looks good to merge.

source/Session.cpp
jspjutNV: I agree it should never be hit, but there's probably a better fallback. I'll revise.

bboudaoud-nv: This is likely just fine. I guess the `else` case from the `if` above (`1.0f / float(m_app->window()->settings().refreshRate)`) could be this default...

bboudaoud-nv (approved):

jspjutNV:
I just tried building and running this, and I get a memory fault at the end of creating the results file function that I don't understand. I'm not sure when this regression showed up, but it seems similar to what I've been experiencing with #314. It would be good to know if this is only happening on my machine, or if others also see this bug.

jspjutNV:
I believe I've gotten it to build by changing line 182 in `Logger.cpp` to make the first `"'"` literal into a `G3D::String` with the following.

```c++
		String("'") + sessConfig->id + "'",
```

Since there's a bunch more of the same implicit conversion to G3D strings to get the G3D `+` operators to work, it seems like we should probably be explicitly making those strings into `G3D::String` types. I cannot explain why this only breaks on a couple branches, some of which were working for me before, but doesn't break on the master branch.

bboudaoud-nv:
@jspjutNV can you confirm that the `sessConfig->id` was valid when the crash occurred? We use the literal + G3D::String combo throughout the logger and have never seen any issues from this.

jspjutNV:
> @jspjutNV can you confirm that the `sessConfig->id` was valid when the crash occurred? We use the "[literal] + `G3D::String`" combo throughout the logger and have never seen any issues from this.

I cannot because it isn't crashing any more so I can't inspect to see what's wrong.

jspjutNV:
I ran some tests with a variety of sequences, and the results match my expectations. Sharing the plots of those tests here.

image
image
image

PR #321, opened 2021-09-22
Add `referenceTargetHitSound` to audio config

bboudaoud-nv:
This branch adds support for a `referenceTargetHitSound` config parameter (along with a `referenceTargetHitSoundVol` parameter for volume control) to the audio configuration within the FPSci general configuration.

To demonstrate you can add:
```
referenceTargetHitSound = "sound/fpsci_destroy_150ms.wav";
```
to (the top of/any session in) your experiment configuration file. This should automatically play the specified destroy sound for the reference target. Currently the behavior in this branch defaults to no audio for the reference target, but we can change this default value if we'd like.

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -190,10 +190,15 @@ For more information on G3D `Sampler` options refer to [this reference page](htt
 |-----------------------|-------|-----------------------------------------------------------------------------------------------------------|
 |`sceneHitSound`        |file   | The sound to play when the scene (not the target) is hit by a weapon (for no sound use an empty string)   |
 |`sceneHitSoundVol`     |ratio  | The volume of the scene hit sound to play                                                                 |
+|`referenceTargetHitSound`|file | Sound to play when the reference target is hit (for no sound use an empty string)                         |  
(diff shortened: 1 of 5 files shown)


joohwankimNV:
Feature works great, thank you. But I realize I have an additional request.

While trying it out, I realize the shooting sound is also missing for reference targets. Could we have another parameter that enable/disables shooting sound as well?

bboudaoud-nv:
@joohwankimNV I've added a (boolean) `referenceTargetPlayFireSound` parameter to the general config that allows you to specify whether the weapon's fire sound is played when the reference target is shot at.

This change comes w/ a caveat though...

We only allow single fire (non-automatic/continuous) weapons for the reference target. However, weapon's can be specified with looped fire audio. For now the audio specified for the weapon will _only_ play if it is non-looped. I've documented this in the audio config section of `general_config.md` for now.

Per discussion w/ @jspjutNV we may want to eventually consider moving to a model that allows a different weapon to be specified for the reference target. This would allow things like custom miss decals, fire sounds, and models and eliminate the need for the special-purpose flags implemented in this branch.

jspjutNV (approved):
This PR appears to work as described by the documentation.

PR #322, opened 2021-09-23
Add `clearMissDecalsWithReference` parameter

bboudaoud-nv:
This branch adds support for a `clearMissDecalsWithReference` parameter that allows the experiment designer to specify whether or not miss decals created while firing at the reference target are cleared when the trial task begins. The parameter defaults to `false` to preserve historical behavior. To enable, add:
```
clearMissDecalsWithReference = true;
```
to your experiment config.


joohwankimNV:
Noticing a bud regarding when decals are cleared. I think we want to clear decals together with reference target (at the beginning of pretrial period). Right now the decals are cleared after pretrial period.

jspjutNV:
> Noticing a bug regarding when decals are cleared. I think we want to clear decals together with reference target (at the beginning of pretrial period). Right now the decals are cleared after pretrial period.

The latest commit should have fixed this to match your description. If you have a chance to try it again, make sure it's working how you expect and we can merge this PR.

jspjutNV (approved):

PR #323, opened 2021-09-23
Add `showReferenceTargetMissDecals` parameter

bboudaoud-nv:
This branch adds a `showReferenceTargetMissDecals` parameter to allow experiment designers to specify whether miss decals should be displayed when firing at the reference target. This parameter defaults to `true` to maintain historical behavior. To disable it, add:
```
showReferenceTargetMissDecals = false;
```
to your experiment config.

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -433,6 +433,7 @@ These flags help control the behavior of click-to-photon monitoring in applicati
 |`referenceTargetColor` |`Color3`               | The color of the "reference" targets spawned between trials                        |
 |`referenceTargetSize`  |m                      | The size of the "reference" targets spawned between trials                         |
 |`showPreviewTargetsWithReference` |`bool`      | Show a preview of the trial targets (unhittable) with the reference target. Make these targets hittable once the reference is destroyed |
+|`showReferenceTargetMissDecals`|`bool`         | Show miss decals when the weapon is firing at a reference target?                  |
 |`previewTargetColor`   |`Color3`               | Set the color to draw the preview targets with (before they are active)            |
 
 ```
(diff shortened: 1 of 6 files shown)


joohwankimNV:
The feature is good. I am noticing a bug, though.

As I switch sessions, going from `showReferenceTargetMissDecals = false` to `showReferenceTargetMissDecals = true` seems to not work. The other way works.

jspjutNV:
> The feature is good. I am noticing a bug, though.
> 
> As I switch sessions, going from `showReferenceTargetMissDecals = false` to `showReferenceTargetMissDecals = true` seems to not work. The other way works.

I can confirm this bug reproduces.

I also noticed a bug that's been in there for a while where there's a segmentation fault on switching sessions from the GUI with certain decals present. I'm not entirely sure which situations cause it because it only happens occasionally.

jspjutNV:
I also noticed a bug that's been in there for a while where there's a segmentation fault on switching sessions from the GUI with certain decals present. I'm not entirely sure which situations cause it because it only happens occasionally. This crash happens in the following code from `Weapon.cpp` on the line with `m_scene->remove(m_currentMissDecals.pop());`:
```c++
void Weapon::clearDecals(bool clearHitDecal) {
	while (m_currentMissDecals.size() > 0) {				// Remove and clear miss decals
		m_scene->remove(m_currentMissDecals.pop());
	}
	m_missDecalTimesRemaining.clear();						// Clear miss decal timeouts

	if (clearHitDecal && notNull(m_hitDecal)) {				// Clear hit decal (if one is present)
		m_scene->remove(m_hitDecal);
	}
}
```

jspjutNV:
I'm going to open an issue for the above problem (if there isn't one yet), then this PR should be ready to merge.

PR #324, opened 2021-09-23
Turn off sounds for test experiment

jspjutNV:
This pull request disables all sounds that should play in the current set of test cases. This should make running the tests a bit more bearable since they'll be quiet.

I did intentionally leave a sound on for the target that shouldn't be destroyed in case it does it get destroyed since sound when the test case fails should be fine.

### DIFF
--- a/data-files/test/experimentconfig.Any
+++ b/data-files/test/experimentconfig.Any
@@ -24,6 +24,7 @@
         instantKill = true;
         damagePerSecond = 100;
         autoFire = false;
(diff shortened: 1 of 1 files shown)


jspjutNV:
In an effort to get ready for Milestone v21.09.01 I'm going to close this one myself. If anyone finds any issues with it after the fact, we can re-open to fix them, but I can't imagine turning off sounds in the test cases will cause any problems.

PR #329, opened 2021-09-28
Trial invalidation from pretrial aim displacement

### LINKED ISSUE #317: Azimuth sign flip between target spawn and player aim
It seems that the azimuth value reported in the `spawn_ecc_h` in the `Targets` table has the opposite sign from the azimuth value reported in the `Player_Action` table's `position_az` column. We should investigate this math and make sure the two azimuths are computed the same way.

### LINKED ISSUE #325: Invalidate trial initiation based on player mouse displacement
Description of the requested feature:
Decide whether to start a trial or not based on aiming direction during the Pretrial period. If the aiming direction stays within the designated angular area, move onto the next state to start the trial. If the aiming direction moves out of the designated angular area, halt the progression of trial states and print a warning text with a warning sound (specified by the experimenter).

There can be multiple ways to implement invalidation process. The first use case of this feature will be to invalidate if aiming direction moves away from the reference target, although reference target is no longer visible during Pretrial state. So I propose an implementation idea of defining an invisible `validationTarget` and use that to validate or invalidate trial progression.

Expected new parameters and algorithm based on that proposal of implementation:

Configuration parameters:
`validationTarget`: Target to be used by the validation algorithm. Most likely to be the same as reference target's definition but invisible. We can force it to be static and invisible, but we can also be flexible to leave them as free parameters and let the experimenter control it. We do want to provide a color option or boolean parameter to make the target invisible though.
`invalidationFeedback`: Message to be printed out when the given trial is invalidated.
`invalidationSound`: Sound to be played when the given trial is invalidated.
`invalidationSoundVol`: The volume of the invalidation sound to play.

Description of validation algorithm:
Continue to check if the aiming direction ray intersects validation target. Invalidate trial initiation if the aiming ray does not intersect validation target in any frame during `PretrialDuration`.

bboudaoud-nv:
This branch adds support for invalidation of trials wherein the player's aim exceeds a provided `maxPretrialAimDisplacement` in degrees (which defaults to 0, implying no max pretrial aim displacement). Upon invalidating a pretrial aim direction for exceeding the threshold a `pretrialAimInvalidFeedback` message is shown the user and the trial is restarted without advancing an trial tracking variables.

As a side effect this PR addresses the issues discussed in #317, wherein the scene config's `spawnHeading` was ignored and targets where spawned based on an `eccH` value that disagreed in sign with the player's recorded view direction azimuth. In the unified convention positive angle refers to clockwise rotation (or rotation that turns the view to the right).

Merging this PR closes #317 and closes #325


bboudaoud-nv:
As part of testing this PR I've noticed that clearing the `m_feedbackMessage` string at line 473 of `Session.cpp` seems to be creating an assertion issue when running in the `Debug` config. This is particularly puzzling as the routine throwing the exception seems to be checking for proper memory allocation (the string is always over-allocated as we are clearing it to its minimum size). 

Use of `my_str = ""` and `my_str.clear()` both appear to be creating exceptions here.

bboudaoud-nv:
We've now added the current task state to both the `Player_Action` and `Target_Trajectory` tables and log to these tables in all presentation states (not just `trialTask` as previously). 

To help reduce data file size for long running experiments we've added support for a `logOnChange` parameter to the logger that avoids logging duplicate entries to the `Player_Action` and `Target_Trajectory` tables when a target/player doesn't move at all.

docs/general_config.md
```diff
@@ -73,7 +74,8 @@ The following settings allow the user to control various timings/durations aroun
 "trialFeedbackDuration": 1.0,               // Time for user feedback between trials
 "sessionFeedbackDuration": 5.0,             // Time for user feedback between sessions
 "sessionFeedbackRequireClick" : false,      // Don't require a click to move past the scoreboard
-"defaultTrialCount" : 5,
+"defaultTrialCount" : 5,                    
+"maxPretrialAimDisplacement" : 0,           // Disable max pretrial aim displacement by default
```
jspjutNV, line 78: It's weird that 0, a value you'd naturally want to be able to put in to restrict to 0 motion, does the exact opposite and allows infinite motion. Might be better to use `-1` or any less than 0 value for this purpose, though that is also not quite sensible. `INF` if it is writable might be better.

bboudaoud-nv: We can use a negative value to indicate this is off (not sure on Any serialization of infinity or nan). But, on related note, I would **highly** suggest against using a 0 value here, as it would mean if the mouse is aimed _any_ direction except identically at the 0 ray the trial would be invalid. So if the reference target has a size (which it must), you could hit the reference, but still almost always have an invalid trial. Even in the 1 pixel reference target case this could include possible floating point errors w/ representing 0 exactly.

Makes me think that we might want to include more in the documentation on suggestions for this value...

jspjutNV: What is the 0 direction then? Is that not where the user happened to aim to destroy the target? I was assuming this limit was based on the aim direction when the target was destroyed, not some other 0 direction.

bboudaoud-nv: The 0 direction is the direction through the _exact center_ of the reference target (the canonical az/el = 0 direction up to the scene `spawnHeading` specified in a config). Another way to think of it is the 0 `eccH`/`eccV` value used to spawn task targets. 

Using this "true" 0 as opposed to where the player clicked on the reference avoids players "cheating" towards and edge of a reference target, then cheating again with pretrial movement. It also helps make task ID more uniform across trials.

jspjutNV: I see how a 0 entry wouldn't be useful. I do think it's a weird value to mean infinity. On a related note, would 180 or 360, or maybe both give the same behavior?

bboudaoud-nv: I hadn't thought about large values here, but I suppose allowing up to 180° offset would basically give the full 360° view rotation as valid. This could also be an option.

I'm open to suggestions, but as implemented this value must be >0° for the pretrial aim displacement test to be applied.

joohwankimNV: I like 180 more, which makes sense in physical sense. I can also imagine naive experimenters associating 0 with 'no displacement is allowed'. The 180 option will prevent that confusion.

bboudaoud-nv: I realize that the way this angle is computed the aim displacement can exceed 180°. The math to decide on aim displacement is:

```
float viewDisplacement = sqrt(pow(az_displacement, 2) + (el_displacement,2))
```

So, if you produce an azimuth displacement of 180° (largest possible) then _any_ elevation displacement the resulting angle will be greater than a 180° threshold. I suppose 360° would be safe here instead, though maybe we want to calculate this angle differently.

jspjutNV: I'd suggest we change the math to make it a circular conic solid angle of the provided max displacement rather than using the current square cone. In other words, the orange in this diagram rather than red. 
image

- Image from [here](https://www.solarxy.org/2020/11/17/solar-radiation-the-fundamentals/)

bboudaoud-nv: We were trying to use a circle. but may have done so with incorrect assumptions. 

The idea from  the code above is that the view displacement is given by the length/magnitude of the angular vector (i.e. [azim, elev]) pointing from the "ideal 0" view direction to the actual view direction. By thresholding the magnitude of this difference vector we establish a circular region in angle space over which we will allow view displacement. Thinking about this more I think its an incorrect approach to measuring the angle between two vectors... I think the usual approach is taking the arccos of the normalized dot product to get the angle between two vectors. I will look into moving to this approach shortly.

I believe the square region comes from thresholding both azimuth and elevation independently (which we are not doing). Having said that I think we may

jspjutNV (approved):
Functions as I'd expect. I didn't double check the target and player heading agreement, but we may want to build a test case to ensure that correct behavior persists in the future.

PR #331, opened 2021-10-04
Add `logSystemInfo` parameter and USB VID/PID lookup

### LINKED ISSUE #288: Collect information about current mouse
This is a feature to collect information about the mouse being used and store it in the database. We'd like the following:

* Mouse make and model
* Model variant (if any)
* Firmware version
* Polling rate
* Click latency
* Sensor latency
* Mouse DPI (may not be possible)

bboudaoud-nv:
This branch adds support for logging `SystemInfo` to the output database as a new `System_Info` table. It also adds support for getting the VID/PID of all USB devices attached to the system and logging this as part of `SystemInfo` (which is also written to `log.txt` as it has been historically).

In order to get VID/PID this branch adds a dependency on `setupapi.lib` which will now need to be packaged with the project. This is currently a windows-only feature.

There is no support built-in for identifying mice/keyboards from other USB devices, but the VID/PID of every connected device is reported. An optional `vid_pid.json` file has been added to the `data-files` directory to provide a lookup of vendor/product names from the VID/PID pair. This file is produced using a [Python web-scraping tool](https://colab.research.google.com/drive/1tRE-ErkH7NAlTYjpu6vExyvypAGNtqzk?usp=sharing).

Merging this PR closes #288.


jspjutNV:
Is `setupapi.lib` a G3D project? This is a great idea, but we need to make sure we can package the library before we can merge this one.

PR #334, opened 2021-10-05
Scene heading

### LINKED ISSUE #332: Simplify scene-based player/camera heading
Currently the approach to getting the heading of a `PlayerEntity` from a scene.Any file (which occurs when no `spawnHeading` is specified in the experiment config `scene` specification) is quite convoluted and confusing. We should simplify it.

- Per discussion with @jspjutNV we should simplify this to allow more direct/uniform specification of the player frame within a scene using the `frame = CFrame::from...()` style specifications used elsewhere in G3D scene files.
- We should eliminate the `heading` value from the `PlayerEntity` Any serialization as this is no longer needed
- We should always default to the frame specified in the `PlayerEntity` over the player/default camera and make sure that when the player camera is created its frame is inherited from the `PlayerEntity` in code

Currently this issue can be demonstrated by loading the default scene (`Simple_Hallway.Scene.Any`) and observing that regardless of the `PlayerEntity` `heading` value the `position_az/el` reported in the `Player_Action` table appears to be inherited from the `defaultCamera` specification and not the `PlayerEntity`. This should be corrected such that no initial frame should need to be provided for cameras in FPSci scene specifications.

bboudaoud-nv:
This branch cleans up scene-level specification of `PlayerEntity` heading using the `frame` of the `VisibleEntity` as opposed to a purpose-built Any field. It also attempts to fully unify the player camera and player heading from when the `PlayerEntity` is loaded from Any.

Merging this PR closes #332

### DIFF
--- a/data-files/scene/Simple_Hallway.Scene.Any
+++ b/data-files/scene/Simple_Hallway.Scene.Any
@@ -17,8 +17,7 @@
         
 		player = PlayerEntity {
             model = "playerModel";
-            frame = Point3(47.522999, -2.3549, -0.35916	);
-			heading = -1.5707963267948966192313216916398;
+            frame = CFrame::fromXYZYPRDegrees(47.523, -2.3549, -0.35916, -90, 0, 0 ); 
         };
 		
 		camera = Camera {
@@ -82,7 +81,6 @@
                 vignetteTopStrength = 0.5; 
             };
             
-            frame = CFrame::fromXYZYPRDegrees(47.523, -2.3549, -0.35916, 87.025, 3.163, 0 ); 
             mass = 1; 
             motionBlurSettings = MotionBlurSettings { 
                 enabled = false; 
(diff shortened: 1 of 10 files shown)


data-files/scene/Simple_Hallway.Scene.Any
```diff
@@ -17,8 +17,11 @@
         
 		player = PlayerEntity {
             model = "playerModel";
-            frame = Point3(47.522999, -2.3549, -0.35916	);
-			heading = -1.5707963267948966192313216916398;
+            // When the player entity is specified, its frame overrides the camera frame
+            // even if not specified and filled with the default.
+            // If you prefer to use a specified camera frame, you can specify which camera to use
+            // in the experiment config.
+            frame = CFrame::fromXYZYPRDegrees(46.08, -2.2, 0.0, -90, 0, 0 );
```
jspjutNV, line 24: It seems to me that `46.0` in the x is more likely to be centered than `46.08` like it is now.

bboudaoud-nv: If you think this is correct I'm fine w/ changing it. Since we are changing the spawn position from past behavior here, we might as well pick the most "correct" option possible. I do not have strong opinions on this spawn position, and we can always refer to older scene configs to recover the historical spawn position.

PR #342, opened 2021-10-19
Initialize lastADCTime to fix error

bboudaoud-nv:
This branch adds a fix for initializing the `lastADCTime` variable at the top of the file to prevent an exception that prevents the `event_logger.py` script from running.


jspjutNV (approved):
Looks reasonable.

PR #345, opened 2021-11-09
Remove emergencyTurbo mode

bboudaoud-nv:
This branch removes `emergencyTurbo` which was causing issues with controlling AA, bloom, and AO (from the scene) when not set.


jspjutNV (approved):
Great change! Gets rid of dead and buggy code!

PR #346, opened 2021-11-17
Improve Question Specification

### LINKED ISSUE #344: Question Improvement
The current per-session question system in FPSci is functional, but leaves much to be desired in clarity of configuration and flexibility. There are a number of things we could improve around this functionality including:

- Allowing questions to be presented on a per-trial basis in addition to per-session
- Improve configuration
    - The difference between `MultipleChoice` and `Rating` questions is unintuitive (multirow vs single row button layout)
        - When using `MultipleChoice` the number of buttons per row is fixed, this could be made programmable
    - The `randomOrder` parameter defaults to true for all types of questions (unintuitive especially for `Rating`)
    - We could support a drop-down style question in the future

Suggested immediate changes include:
- Make `randomOrder` default to `false` when the question `type` is `"Rating"`
- Add a `buttonsPerRow` that defaults to `3` for `MultipleChoice` and the length of the `options` array for `Rating`
- Consider uniting `MultipleChoice` and `Rating` questions given the availability of `buttonsPerRow` option above (this parameter would always default to the length of the `options` array)

Longer term changes include:
- Add support for per-trial questions (and logging these results affiliated with the trial)
- Add support for drop-down questions
- Add support for multi-line entries in the `options` array (likely using a `\n` character)

bboudaoud-nv:
This branch improves specification of questions to clarify some issues and improve control of presentation of questions. Changes include:

1. Automatically default `randomOrder` to false when displaying questions with `type` of `Rating`
2. Add a `optionsPerRow` field that defaults to `3` (maintains old behavior) but allows custom layout of questions with `type` of `MultipleChoice`
3. Warn the user (via log.txt) that when a question with `type` of `Rating` is provided the `optionsPerRow` will be ignored in favor of displaying all options on one row (use `MultipleChoice` questions for controlled layout).
4. Allows font size setting for questions.

Merging this PR closes #344

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -297,7 +297,7 @@ Questions are configured on a per experiment/session basis using the `questions`
 
 | Parameter Name        | Units  | Description                                                                      |
 |-----------------------|--------|----------------------------------------------------------------------------------|
-|`type`                 |`String`| The question type (required), can be `"MultipleChoice"`, `Rating`, or (text) `"Entry"`      |
+|`type`                 |`String`| The question type (required), can be `"MultipleChoice"`, `Rating`, `DropDown` or (text) `"Entry"`      |
 |`prompt`               |`String`| The question prompt (required), a string to present the user with                |
 |`title`                |`String`| The title for the feedback prompt                                                |
 |`options`              |`Array<String>`| An array of `String` options for `MultipleChoice` questions only          |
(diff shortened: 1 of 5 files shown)


jspjutNV (approved):
Looks good to me, and worked fine in my quick tests.

PR #347, opened 2022-02-02
SIGGRAPH Asia 2019 Sample Experiments

jspjutNV:
This PR includes sample experiments originally used for the [SIGGRAPH Asia 2019 paper](https://research.nvidia.com/publication/2019-11_Latency-of-30) updated to use the latest versions of FPSci.

It also includes updates to the sample experiment contribution instructions.


bboudaoud-nv:
I have made some minor corrections to whitespace formatting and specified the scene name (as you always should when applying position/heading).

However, I think there are still significant differences between this configuration and what was used for the SIGA 2019 study. I've included our screenshot from the SIGA paper below for reference

**Original**
image

**Current Version**
image

Note that:
1. We are not using the same type of cooldown indicator (may be okay)
2. The click-to-photon monitor appears to be on the wrong side (left)
3. The click-to-photon monitor is 2 drastically different sizes between the experiments (track shown in screenshot above seems better)
4. The reticle is a different color (green vs red)
5. The reticle seems substantially larger (by factor of 2x or more)
6. The reference target seems marginally smaller

bboudaoud-nv:
In addition the config file itself seems to be using a mix of syntax that might be confusing for new users to FPSci, this includes:

1. Use of `,` and `;` to end lines
2. Use of `=` and `:` in assignment
3. Use of `(...)` and `[...]` list conventions
4. Parameter names with and without quotes around them

bboudaoud-nv:
@joohwankimNV overall these changes are looking really good. My only other suggestion is that if we want to we could introduce an `sa2019_sampleuser` that does reticle scale/color rather than updating the default user. Alternatively we could just keep with your approach of updating the global `Sample User`. 

You could add:

```
  reticleColor = [Color4(0,1,0,1), Color4(0,1,0,1)];
```

to the user config to make the reticle green if you'd like. I'm not sure how important target/reticle contrast is overall, but I don't see a downside to more closely matching the experiment.

jspjutNV:
I'm going to merge this despite not being "approved" since the only remaining issue was fixed by Joohwan and we really need to get the release out. If you want to post-approve it still Ben, feel free.

PR #350, opened 2022-02-08
Experiment Configured Reticle

### LINKED ISSUE #349: Experiment specified experiment configurations
A session and experiment should be able to set a reticle setting which (when specified) will override the user settings.

In implementing this, it may be a good opportunity to create a "preferences" class to specify various per-user preferences and have the experiment override them when it's a part of the experiment design.

bboudaoud-nv:
This branch adds support for an experiment specified reticle and generally adds some clean-up around previously user-specified only values that can now be configured as part of an experiment configuration (as independent variables).

The behavior is that user-specified reticle behavior is overridden by experiment/session level configuration for these parameters whenever it is provided. This could be extended to include mouse sensitivity parameters as well.

Closes #349


bboudaoud-nv:
Currently the concept of users changing reticle/sensitivity in the user menu is decoupled from the behavior in app. Perhaps we should automatically set the fields allowing this change in the config to false if we detect an experiment-provided config?

Also logging still needs to be updated to be accurate for cases in which this configuration comes from the experiment. Currently this information is only logged to the `Users` table.

jspjutNV:
If you include `sessionParametersToLog = ("reticleIndex", "reticleScale", "reticleColor", "reticleChangeTime");` in the experiment, then the set values will be logged with each session. I think this is a decent work around, and I'm fine with it also being logged in the Users table. 

I also don't mind having a settings combination that allows users to think they're changing the reticle, but having those changes prevented by the experiment or session.

jspjutNV (approved):

PR #357, opened 2022-04-07
Random order question key bind fix

### LINKED ISSUE #356: Issue with randomOrder questions w/ keyboard responses
When answering questions with `randomOrder = true` using key-bound responses the answers reported in the results file do not match the user selection.

To reproduce create any multiple choice style question with `randomOrder = true` and `optionKeys` specified. Note that the randomization of options may result in the original ordering, but whenever it does not the recorded response does not indicate the correct selection.

bboudaoud-nv:
This branch fixes a bug when using `randomOrder` questions with `optionKeys` specified. Previously logged responses were potentially incorrect in this configuration. Now the mapping of `options` to `optionKeys` does not change when `randomOrder = true` is specified in config. Only the presented order of options is changed (the key binds remain the same).

Merging this PR closes #356.

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -313,6 +313,8 @@ Questions are configured on a per experiment/session basis using the `questions`
 
 The user can specify one or more questions using the `questions` array, as demonstrated below.
 
+Note that when using `randomOrder` the options key bindings specified in `optionKeys` remains static (i.e. while ordering of presented options will change the keypress used to enter any given option will remain constant between the `options` and `optionKeys` array).
+
 ```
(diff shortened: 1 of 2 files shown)


bboudaoud-nv:
It is worth mentioning that the current fix uses a nested for loop to find the correct index for the option key bind between the (possibly randomized) dialog options and original question options arrays. Alternatively we could use a table-based lookup for these key binds instead, improving runtime by making question configuration serialization a bit more complicated. This may be worth looking into.

jspjutNV (approved):
Looks good

PR #359, opened 2022-04-11
Add specification of target glossiness to configuration

bboudaoud-nv:
This branch reverts a hard-coded glossiness assumption that was previously used in FPSci. Targets and session/experiment level config can now specify their `gloss` and `targetGloss` respectively with per-target gloss overriding the global settings (similar to target color configuration).


jspjutNV:
In an effort to document how to retain the old behavior, I'm going to make this comment, and update it to have the right config settings to keep the old behavior. I'll put this comment into the release notes.

In each target, add the following:
```c++
gloss = Color4(0.4f, 0.2f, 0.1f, 0.8f);
```

jspjutNV (approved):

PR #362, opened 2022-04-11
Add Sphere models

jspjutNV:
This PR adds low and mid poly sphere models created by @bboudaoud-nv . It also adds a sample experiment to demonstrate them as well as the available and documented high poly sphere that can be [downloaded from here](https://drive.google.com/file/d/1LvJaJUD3k7DR0taZYZ_9Y0PNVQDMdShM/view?usp=sharing).


jspjutNV:
I'm going to try changing the default target model to the low polygon sphere instead of the icosohedron. Let's keep this open until we're happy with this change.

jspjutNV:
This is now ready for review.

PR #364, opened 2022-04-13
Remove duplicate save of user status

bboudaoud-nv:
This branch removes a 2nd save of user status from the session state machine.

### DIFF
--- a/source/Session.cpp
+++ b/source/Session.cpp
@@ -505,7 +505,6 @@ void Session::updatePresentationState()
         
 				// Save current user config and status
 				m_app->saveUserConfig(true);											
-				m_app->saveUserStatus();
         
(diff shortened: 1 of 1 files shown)


bboudaoud-nv:
There were previously 2 calls made to `m_app->saveUserStatus()` in `Session.cpp`. This branch removes the call from `Session::updatePresentationState()` in favor of the call in `FPSciApp::markSessionComplete()` (also called from `Session::updatePresentationState()`. This makes sense as it keeps the `saveUserStatus()` call contained inside of the `FPSciApp` and avoids the duplicate call we had previously.

jspjutNV:
I found the other save that we're keeping. It's in `m_app->markSessComplete` which is called in the `trialFeedback` state right before it transitions to the `sessionFeedback` state.

jspjutNV (approved):

Issue #348, opened 2022-02-02
Experiment and user selection interaction


bboudaoud-nv:
When running in developer mode the experiment selection and user selection process appears to interact problematically from the in-app menu.

Essentially the ability to select a different experiment from the experiment drop-down in the user menu without "loading" it in the sense of pressing the "Select Experiment" button in the menu allows for users to accidentally overwrite another experiment's user status with the status from the currently loaded one.

You can reproduce this issue by:

1. Open FPSci in developer mode (`developerMode = true` in startup config)
2. Select a different experiment (w/ different session names) from the experiment drop-down menu but DO NOT click the "Select Experiment" button
3. Press the "Select User" button to select the current user (or another one)
4. Now press the "Select Experiment" button to produce an error (session ID from user status missing from experiment)

You will also notice that now the user status file for the experiment that was selected in the drop down, but not "loaded" in the sense of pressing the "Select Experiment" button, has been written with the originally loaded experiment's user status, breaking loading this experiment on all future runs of FPSci.

bboudaoud-nv:
For now you should attempt to avoid this state by always pressing the "Select Experiment" button prior to taking another action from the user menu.

In the event the user status file is overwritten by this bug, you may be able to revert the changes if the overwritten user status is checked into git. Otherwise you will need to rely on a local/alternative backup to restore your user config.

PR #365, opened 2022-04-18
User session status split

### LINKED ISSUE #355: Performance degredation when user status has a large number of users
We've had a report that FPSci appears to slow down significantly when more than 50 users are listed in the user status (and config) file. Someone will need to investigate this problem.

As a work around, you can remove participants from the file who aren't needed.

bboudaoud-nv:
This branch splits the per-session written user status (sessions completed) from the remainder of the user status file. The new per-session status file format supports an iterative write approach and therefore serializes to disk faster. This file uses a CSV format with `user ID`, `session ID` formatting for each line. It is read in once from disk at startup to recover the user session status for all users.

The old user status file still contains the `currentUser` field (and potentially adding new users) which requires write-back when the user changes/is added but otherwise should be written far less frequently.

Custom session status log names can be specified using the new `completedLogFilename` configuration parameter in the top-level of the user status file.

Merging this branch closes #355


bboudaoud-nv:
This might interact w/ #348, perhaps we should attempt a fix for that issue in this branch

jspjutNV:
I checked out this branch and it fails to build. I'm not sure what's going on yet, but the error happens outside of the code from this PR, but doesn't happen on the master branch. I'm investigating.

jspjutNV:
73f8b05 fixed the build error.

jspjutNV:
There appears to be a remaining bug, or at least weird behavior if `allowRepeat` is on, and the user selects sessions out of order. We sort of cover this in the readme, and `allowRepeat` is intended to be only used when the user can't select session order, so I think we can consider this documented behavior. I'll clarify a little in the docs.

jspjutNV (approved):
I made some changes to docs and fixed the test case that was no longer testing everything. This appears to work in my tests, but we'll want to be aware that something may break and need fixing. This does break the current "create user" hack somewhat used by some experiments, so we will want to prioritize that fix for the next release.

PR #366, opened 2022-04-19
Check for legal results filename

### LINKED ISSUE #351: Problem with slash in experiment description
Putting a `/` (or any character incompatible with a filename) in the experiment `description` parameter will break results output without an error as the experiment description is used in the results filename.

This string is also used in the session's table `appendingDescription` column, but should not be an issue here.

bboudaoud-nv:
This branch adds a `FilePath::isLegalFilename()` check on the filename used for the results database. Previously the results could fail to write due to an illegal character in either an experiment description or a user name.

The new check should throw an exception when any invalid filename character is detected in the results filename. It is worth noting this may catch characters that technically legal but not encouraged for file naming.

Merging this PR closes #351

### DIFF
--- a/source/FPSciApp.cpp
+++ b/source/FPSciApp.cpp
@@ -722,10 +722,15 @@ void FPSciApp::updateSession(const String& id, bool forceReload) {
 		FileSystem::createDirectory(resultsDirPath);
 	}
 
-	const String logName = sessConfig->logger.logToSingleDb ? 
-		resultsDirPath + experimentConfig.description + "_" + userStatusTable.currentUser + "_" + m_expConfigHash :
-		resultsDirPath + id + "_" + userStatusTable.currentUser + "_" + String(FPSciLogger::genFileTimestamp());
+	// Create and check log file name
+	const String logFilename = sessConfig->logger.logToSingleDb ? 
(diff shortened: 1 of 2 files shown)


bboudaoud-nv:
Currently this branch also introduces an issue with descriptions/user names that include spaces. This should be resolved before merging...

bboudaoud-nv:
For now I've moved to using `FilePath::makeLegalFilename()` as a quick fix here. This is a work-around for the problem of temporary user names (i.e., "Create a new user") and experiment descriptions with spaces in them triggering a runtime exception.

This may only cause issues when something like 2 user IDs/experiment descriptions only differ by sequential illegal characters in which case they may collide... I think this branch is a fine fix for the issue for now

jspjutNV (approved):
All tests passed and this looks like a solid, straightforward improvement.

jspjutNV:
Actually, it looks like there's a likely bug with an extra `+ ".db"` in there. Checking that before merging.

jspjutNV:
> Actually, it looks like there's a likely bug with an extra `+ ".db"` in there. Checking that before merging.

This should be fixed with 137b0da

PR #367, opened 2022-04-21
Questions to Any

bboudaoud-nv:
Questions are now serialized to Any, this is importantly primarily so that they are captured in/impact hashes appended to the results file.


bboudaoud-nv:
Before merging this PR we should use it as an opportunity to make sure all experiment/general config is correctly written to Any and therefore impacts the hash of the experiment results file.

PR #369, opened 2022-05-03
Target emissive control

### LINKED ISSUE #370: Target color/gloss/emissive ignores session-level config
Currently session-level target color/gloss/emissive parameters are ignored in favor of experiment-level values. We should either:

1. Document this behavior
2. Implement a fix that lets session-level configuration override experiment-wide configuration for these values

bboudaoud-nv:
This branch adds support for setting target emissive color (previously hard-coded to 0.7*color) using 2 different mechanisms. Targets can specify it using an `emissive` parameter or it can be specified at the experiment-level using `targetEmissive`, both of which take a `Color4` as input.

Merging this PR closes #370


jspjutNV (approved):
Tests passed

Issue #275, opened 2021-02-18
The first, or near first test run sometimes fails


jspjutNV:
I've noticed when running a subset of the tests that there's often a failure in various weapon fire tests. This bug may be related to the target sound lists added in #268 .

While this is a problem near the start of a new target sound, in steady state it is unlikely to cause long term issues. Still, we should fix this bug to get all of the tests passing consistently again.

PR #376, opened 2022-05-27
Pre-play audio on first load

### LINKED ISSUE #375: Performance hitch early in runtime
There is a known issue w/ a performance hitch around shooting at/eliminating the first target in a run of FPSci. We believe this may be related to how audio is loaded/played for the first time but have not confirmed this. Currently the issue is not (typically) a problem as the first target shot at/destroyed is typically a reference and thus the bug does not impact "true" data collection.

jspjutNV:
This PR makes audio play when first loaded to avoid stuttering when it first needs to play.

When finished, closes #375 

With some luck may also help #275 , but it doesn't fix that yet.

### DIFF
--- a/source/FPSciApp.cpp
+++ b/source/FPSciApp.cpp
@@ -711,9 +711,13 @@ void FPSciApp::updateSession(const String& id, bool forceReload) {
 	weapon->loadSounds();
 	if (!sessConfig->audio.sceneHitSound.empty()) {
 		m_sceneHitSound = Sound::create(System::findDataFile(sessConfig->audio.sceneHitSound));
+		// Play silently to pre-load the sound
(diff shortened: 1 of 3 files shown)


jspjutNV:
After 4622f42 #275 still isn't fixed and the click and hold test still fails. We might need to push the audio loads and plays earlier, which will require a bit more work.

bboudaoud-nv (approved):
This fix is as good as any for the current model of loading sounds in FPSci. However adding sounds to the `soundTable` in the `set[X]Sound()` methods for the `Target`s is weird. In the future we probably want to build up `soundTable` at config load and just preplay all audio directly from here. This is a fine fix for now.

PR #379, opened 2022-06-07
Add `requireUserAdd` parameter

### LINKED ISSUE #373: UI for creating new user
A relatively minor request.

The steps for creating a new user is,

1. Open the dialog box (by pressing esc)
2. Type in new user name
3. Press the '+' sign next to it

Step 2 was not clear, even though there was an 'Empty!' notification message when I clicked on '+' without typing in the user name. Feels like it can be improved, although it's completely operational so not urgent at all.

bboudaoud-nv:
This branch adds support for a `requireUserAdd` parameter in the general configuration of FPSci. When enabled this flag requires a new user to be added via the existing user menu prior to starting an experiment on each run of FPSci. Note that if FPSci is closed an existing user will still need to create a new user with the current implementation.

The user menu is drawn differently when `requireUserAdd = true`, will always be shown at startup, and cannot be closed until a new user is added in an effort to make the requirement more self-evident. Once a new user is added the typical user menu behavior resumes (allowing experiment configured user customization).

Merging this PR closes #373

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -607,6 +607,7 @@ These flags control the display of the in-game user menu:
 |`showUserSettings`                 |`bool`     |Show the per-user customization (sensitivity, reticle, etc) options    |
 |`allowSessionChange`               |`bool`     |Allow users to change the session using a user menu dropdown           |
 |`allowUserAdd`                     |`bool`     |Allow users to add new users to the experiment                         |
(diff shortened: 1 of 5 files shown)


jspjutNV:
One interesting behavior is that a user is required to be added at startup even if there are existing users who haven't finished, but you can switch to those existing users once you add a new one.

I'm not convinced this is a problem, but we'll want to keep an eye on it in case it causes headaches for designers or users.

jspjutNV (approved):

PR #381, opened 2022-06-15
Add alternate scoring models

bboudaoud-nv:
This branch adds support for some additional scoring models for FPSci. Specifically it adds support for a `scoreModel` and `scoreMultiplier` parameter that allow more customization of scoring. Valid `scoreModel`s include `time remaining` (historical), `targets destroyed`, `shots hit`, `accuracy`, and `trial successes`.

We can add more scoring modes in the future if we would like using this model.


bboudaoud-nv:
If we'd like we could also add any/none of these metrics to the `Sessions` table in the results file.

jspjutNV:
I like the idea of logging these scoring models, but don't want that implementation to hold back this PR. Tracking the outstanding issue in #393 .

jspjutNV (approved):

Issue #390, opened 2022-07-29
Staircase Stimulus


bboudaoud-nv:
This issue is a placeholder for discussion of [staircase stimulus](https://www.psychopy.org/coder/tutorial2.html) in FPSci.

Issues to address include:

1. How to specify the parameter, step size, and maximum step count for the staircase?
2. How to limit invalid configuration caused by experimenter misunderstanding of parameters (i.e. min step < controllable step, parameter cannot be staircased)?
3. How to incorporate into existing FPSci configuration (different session type)?
4. How to manage bugs that might arise in the field but not be present in pilots (long run time/invalid results)?
5. How to implement/where we can borrow a valid staircase stimulus generator algorithm from?

This is not currently planned for development in FPSci, but _is_ worth discussing. We would likely plan to have an early implementation complete around October 2022.

bboudaoud-nv:
As part of this development I think we will need to add a (set of) FPSci `ConfigParameter` classes that generalize our configuration and provide some meta-level information about each configuration parameter. This will be a fairly major facelift to the configuration system that will occur in its own branch prior to implementation of the staircase-style session.

bboudaoud-nv:
Alternatively to a `ConfigParameter` we can potentially use the `Any` config structure to implement this feature. This is closer to the idea of having per trial configuration, which we could unite with this approach.

PR #392, opened 2022-08-17
Trial-level configuration

### LINKED ISSUE #391: Trial-level Configuration
As a precursor to a possible staircase design as discussed in #390, we are considering adding per-trial configuration to FPSci. This comes w/ several challenges. This is a place to summarize them.

1. There are several `FpsConfig` parameters that need to be relocated out of the generic class. Things like `sessionFeedbackDuration`, `sessionCompleteFeedback`, `showMenuBetweenSessions` and `sessionFeedbackRequireClick` don't really make sense to be specified on a per-trial basis. Having said this `allSessionCompleteFeedback` can still be specified at the session-level for now, so perhaps this is okay?
2. It may make sense to have a new `FpsConfig` type structure that is only inherited from the experiment to the session, so that session-level config can be inherited from the experiment, but does not need to be inherited by the `TrialConfig` as well. This applies in places like (1) where we would still like to support experiment-level specification of parameters that apply to all sessions, but won't use this state for individual trials.
3. The FPSciApp/session state machine will need to be updated to call something like `FPSciApp::updateSession()` on every trial, this should be fairly easily done, but we should think about where/how the management of the trial config is done and any perf impacts this may imply
4. We will need a `trialParamsToLog` similar to `sessParamsToLog` (which is another config parameter like those specified in (1) above). This will be the only mechanism for logging configuration that changes between trials. We will need to update the logger for this accordingly
5. Currently questions can only be asked at the end of a session, we should either move the questions array to the session/experiment level (as specified in (2)) or come up with a model for per-trial questions (which I believe will be needed).

bboudaoud-nv:
This branch adds support for trial-level configuration inherited from the experiment and session in turn.

This is a first-step towards adding staircase designs to FPSci as described in #390.

Merging this PR closes #391.


jspjutNV:
I believe there's a bug in the current version of this PR where the dummy target respawns at a different location after each trial, at least in some configurations. The one I first noticed it in is the `sa2019_1hit` sample.

source/FPSciApp.h
```diff
@@ -33,11 +33,12 @@ class WaypointManager;
 // ready: ready scene that happens before beginning of a task.
 // task: actual task (e.g. instant hit, tracking, projectile, ...)
 // feedback: feedback showing whether task performance was successful or not.
-enum PresentationState { initial, pretrial, trialTask, trialFeedback, sessionFeedback, complete };
+enum PresentationState { initial, referenceTarget, pretrial, trialTask, trialFeedback, sessionFeedback, complete };
 static String presentationStateToString(const PresentationState& state) {
 	String stateStr = "N/A";
 	switch (state) {
 	case initial: stateStr = "initial"; break;
+	case referenceTarget: stateStr = "referenceTarget"; break;
```
jspjutNV, line 41: Leaving a comment about the added state to remind me to ask about it.

bboudaoud-nv:
The bug @jspjutNV mentioned above has now been resolved. Specifically it only occurs when `resetPlayerPositionBetweenTrials` is `false` and the view moves away from the default view direction. Along with this change came some fixes for handling getting the correct spawn direction on the first trial of a session when `resetPlayerPositionBetweenTrials` is `false`.

jspjutNV:
I created [a pre-release](https://github.com/NVlabs/FPSci/releases/tag/v22.11.01) for the current state of this PR. There's still some issues, but this should make it easier for non-developers to test their experiments and report bugs.

jspjutNV:
I'm noticing a bug in the current version that seems to be related to having a scene name set without a `spawnHeading`, expecting the default value, or the one set in the scene for the spawn heading. The result is that the dummy target spawns somewhere unreachable in my test, but it may be somewhere reachable in others. Adding a `spawnHeading` to the config seems to resolve the issue, but is a regression compared to previous versions. We'll want to reflect this change in the documentation and release notes if this change is intentional.

bboudaoud-nv:
@jspjutNV yes, this bug was introduced by my "fix" for the issue when `resetPlayerPositionBetweenTrials` is `false`. I need to think a bit more on all the places the spawn heading might come from and come up with a more robust solution for where to locate the reference target here.

bboudaoud-nv:
I'm now running into issues with specifying the `scene` `name` parameter on a trial-by-trial basis. Will need to look into this configuration interaction.

bboudaoud-nv:
I believe I've now fixed the issues with `scene` `name`s changing between trials. @jspjutNV you should be able to test and see if the previous issue with `spawnHeading` not being specified.

jspjutNV:
The latest commits fix most of the issues we saw before.

I've notice that 2 tests are failing, but haven't investigated the root cause yet. These failing tests are:

- [ ] `FPSciTests.TestAutoFire` which fails with the message "Low damage-per-second with auto-fire but target still died"
- [ ] `FFPSciTests.TestFreshStart` which throws an error on startup (I think it was this one), but when you continue says that a bunch of files did not correctly delete and regenerate. The error messages may be misleading about the root cause here.

bboudaoud-nv:
@jspjutNV both of these issues are now resolved. 

`FPSciTests.TestAutoFire` was failing as it was still attempting to configure the weapon through the session-level config (I changed this to use the new trial-level config which is what takes effect)
`FPSciTests.TestFreshStart` was failing due to the `m_loadedScene.name` not being cleared in `initExperiment()` which meant that `GApp::loadScene()` was being called with an empty string as its argument (as you pointed out this also impacted selection of an experiment through the in-app drop down).

data-files/samples/trials.Experiment.Any
```diff
@@ -0,0 +1,142 @@
...
+                    ids = ( "static"); 
+                    cooldownMode = "box";
+                    weapon = { 
+                        id = "2hit"; 
+                        firePeriod = 0.5;
+                        damagePerSecond = 1.1;
+                        autoFire = false;
+                    }; 
+                }, 
+                {
```
bboudaoud-nv, line 41: This trial looks identical to the one above it, am I missing something?

data-files/samples/trials.Experiment.Any
```diff
@@ -0,0 +1,142 @@
...
+                    ids = ( "static"); 
+                    cooldownMode = "box";
+                    weapon = { 
+                        id = "2hit"; 
+                        firePeriod = 0.1;
+                        damagePerSecond = 5.5;
+                        autoFire = false;
+                    }; 
+                }, 
+                {
```
bboudaoud-nv, line 63: Same as above w/ this trial, seems to repeat the trial before it?

data-files/samples/trials.Experiment.Any
```diff
@@ -0,0 +1,142 @@
...
+        damagePerSecond = 2;
+        autoFire = false;
+    }; 
+    sessions = ( 
+        { 
+            // This session tests out different weapon cooldowns and cooldown indicators while maintaining 2 hits to kill
+            id = "1target_cd"; 
+            description = "1 target cooldown"; 
+            trials = ( 
+                {
```
bboudaoud-nv, line 21: I would give each trial an `id` to demonstrate this new feature

data-files/samples/trials.Experiment.Any
```diff
@@ -0,0 +1,142 @@
...
+    showHUD = true;
+    renderWeaponStatus = true;
+    cooldownMode = "ring";
+    cooldownColor = Color4(0.7,1.0,0.9,0.75);
+            
+    weapon = { 
+        id = "2hit"; 
+        firePeriod = 0.3;
+        damagePerSecond = 2;
+        autoFire = false;
```
bboudaoud-nv, line 13: `autoFire` defaults to `false` but we are setting it here, not an issue but I might remove it from below since its repeated quite a bit.

bboudaoud-nv (commented):
I have changes ready to submit for most of these comments, but let me know if I misunderstood the intent of anything you were trying to do here.

data-files/samples/trials.Experiment.Any
jspjutNV: yep, meant to leave out the cooldown type to fall back to the experiment default.

jspjutNV (approved):
Lots of changes, but I think we got most of the bugs. We'll report and fix any remaining later on.

PR #395, opened 2022-09-03
Parameterized Sprinting Functionality Added

Tokey:
This adds a Sprint Multiplier slider to player controls. MoveRate will be multiplied with Sprint Multiplier to make the player move faster as long as LShift is held.

### DIFF
--- a/source/FpsConfig.h
+++ b/source/FpsConfig.h
@@ -71,7 +71,8 @@ public:
 	Vector2			turnScale = Vector2(1.0f, 1.0f);			///< Player (horizontal/vertical) turn rate scaler
 	Array<bool>		axisLock = { false, false, false };			///< World-space player motion axis lock
 	bool			stillBetweenTrials = false;					///< Disable player motion between trials?
-	bool			resetPositionPerTrial = false;				///< Reset the player's position on a per trial basis (to scene default)
+	bool			resetPositionPerTrial = false;				///< Reset the player's position on a per trial basis (to scene default)'
+	float			sprintMultiplier = 2.0f;					///< Multiplies with players move rate when player is sprinting (defaults to 2x)
 
 	void load(FPSciAnyTableReader reader, int settingsVersion = 1);
 	Any addToAny(Any a, bool forceAll = false) const;
--- a/source/KeyMapping.cpp
+++ b/source/KeyMapping.cpp
@@ -7,13 +7,14 @@ KeyMapping::KeyMapping() {
 	map.set("strafeLeft", Array<GKey>{ (GKey)'a', GKey::LEFT });
 	map.set("moveBackward", Array<GKey>{ (GKey)'s', GKey::DOWN });
 	map.set("strafeRight", Array<GKey>{ (GKey)'d', GKey::RIGHT });
+	map.set("sprint", Array<GKey>{ GKey::LSHIFT });
 	map.set("openMenu", Array<GKey>{ GKey::ESCAPE });
 	map.set("quit", Array<GKey>{ GKey::KP_MINUS, GKey::PAUSE });
 	map.set("crouch", Array<GKey>{ GKey::LCTRL });
 	map.set("jump", Array<GKey>{ GKey::SPACE });
 	map.set("shoot", Array<GKey>{ GKey::LEFT_MOUSE });
 	map.set("scope", Array<GKey>{ GKey::RIGHT_MOUSE});
-	map.set("dummyShoot", Array<GKey>{ GKey::LSHIFT });
+	map.set("dummyShoot", Array<GKey>{ GKey::TAB });
 	map.set("dropWaypoint", Array<GKey>{ (GKey)'q' });
 	map.set("toggleRecording", Array<GKey>{ (GKey)'r' });
 	map.set("toggleRenderWindow", Array<GKey>{ (GKey)'1' });
--- a/source/PlayerEntity.cpp
+++ b/source/PlayerEntity.cpp
@@ -90,8 +90,16 @@ void PlayerEntity::onPose(Array<shared_ptr<Surface> >& surfaceArray) {
 }
 
 void PlayerEntity::updateFromInput(UserInput* ui) {
+	
+	float walkSpeed;
 
-	const float walkSpeed = *moveRate * units::meters() / units::seconds();
+	// Check if player is sprinting or not
+	if (!m_sprinting) {
+		walkSpeed = *moveRate * units::meters() / units::seconds();
+	}
+	else {
+		walkSpeed = *moveRate * *sprintMultiplier * units::meters() / units::seconds();
+	}
 
 	// Get walking speed here (and normalize if necessary)
 	Vector3 linear = Vector3(ui->getX()*moveScale->x, 0, -ui->getY()*moveScale->y);
--- a/source/FPSciApp.cpp
+++ b/source/FPSciApp.cpp
@@ -616,6 +616,7 @@ void FPSciApp::initPlayer(bool setSpawnPosition) {
 
 	// Set player values from session config
 	player->moveRate = &sessConfig->player.moveRate;
+	player->sprintMultiplier = &sessConfig->player.sprintMultiplier;
 	player->moveScale = &sessConfig->player.moveScale;
 	player->axisLock = &sessConfig->player.axisLock;
 	player->jumpVelocity = &sessConfig->player.jumpVelocity;
@@ -1137,12 +1138,20 @@ bool FPSciApp::onEvent(const GEvent& event) {
 				scene()->typedEntity<PlayerEntity>("player")->setJumpPressed(true);
 				foundKey = true;
 			}
+			else if (keyMap.map["sprint"].contains(ksym)) {
+				scene()->typedEntity<PlayerEntity>("player")->setSprintPressed(true);
+				foundKey = true;
+			}
 		}
 	}
 	else if ((event.type == GEventType::KEY_UP)) {
 		if (activeCamera() == playerCamera) {
 			if (keyMap.map["crouch"].contains(ksym)) {
 				scene()->typedEntity<PlayerEntity>("player")->setCrouched(false);
+				foundKey = true;
+			}
+			else if (keyMap.map["sprint"].contains(ksym)) {
+				scene()->typedEntity<PlayerEntity>("player")->setSprintPressed(false);
 				foundKey = true;
 			}
 		}
--- a/source/GuiElements.cpp
+++ b/source/GuiElements.cpp
@@ -175,6 +175,11 @@ PlayerControls::PlayerControls(SessionConfig& config, std::function<void()> expo
 		c->setWidth(width*0.95f);
 	}movePane->endRow();
 	movePane->beginRow(); {
+		auto c = movePane->addNumberBox("Sprint Multiplier", &(config.player.sprintMultiplier), "x", GuiTheme::LINEAR_SLIDER, 1.0f, 10.0f);
+		c->setCaptionWidth(width / 2);
+		c->setWidth(width * 0.95f);
+	}movePane->endRow();
+	movePane->beginRow(); {
 		auto c = movePane->addNumberBox("Jump Velocity", &(config.player.jumpVelocity), "m/s", GuiTheme::LINEAR_SLIDER, 0.0f, 50.0f, 0.1f);
 		c->setCaptionWidth(width / 2);
 		c->setWidth(width*0.95f);
--- a/source/PlayerEntity.h
+++ b/source/PlayerEntity.h
@@ -31,6 +31,8 @@ protected:
 	bool			m_motionEnable = true;				///< Flag to disable player motion
 	bool			m_jumpPressed = false;				///< Indicates whether jump buton was pressed
 
+    bool            m_sprinting = false;                ///< Is the player sprinting?
+
     PlayerEntity() {}
 
 #ifdef G3D_OSX
@@ -50,6 +52,8 @@ public:
 	float*			moveRate = nullptr;	        ///< Player movement rate (m/s)
 	Vector2*		moveScale = nullptr;	    ///< Player X/Y movement scale vector (interpreted as unit vector)
 	Array<bool>*	axisLock = nullptr;		    ///< World-space axis lock
+
+    float*          sprintMultiplier = nullptr; ///< Sprint speed multiplier
 	
 	float*			jumpVelocity = nullptr;		///< Player vertical (+Y) jump velocity
 	float*			jumpInterval = nullptr;		///< Player minimum jump interval limit
@@ -106,7 +110,7 @@ public:
 	void setCrouched(bool crouched) { m_crouched = crouched; };
 	void setJumpPressed(bool pressed=true) { m_jumpPressed = pressed; }
 	void setMoveEnable(bool enabled) { m_motionEnable = enabled; }
-
+    void setSprintPressed(bool sprinting) { m_sprinting = sprinting; };
 	void setRespawnPosition(Point3 pos) { m_respawnPosition = pos; }
     void setRespawnHeadingDegrees(float headingDeg) { m_spawnHeadingRadians = pif() / 180.f * headingDeg; }
 	void setRespawnHeight(float height) { m_respawnHeight = height; }


source/KeyMapping.cpp
```diff
@@ -7,13 +7,14 @@ KeyMapping::KeyMapping() {
...
 	map.set("moveBackward", Array<GKey>{ (GKey)'s', GKey::DOWN });
 	map.set("strafeRight", Array<GKey>{ (GKey)'d', GKey::RIGHT });
+	map.set("sprint", Array<GKey>{ GKey::LSHIFT });
 	map.set("openMenu", Array<GKey>{ GKey::ESCAPE });
 	map.set("quit", Array<GKey>{ GKey::KP_MINUS, GKey::PAUSE });
 	map.set("crouch", Array<GKey>{ GKey::LCTRL });
 	map.set("jump", Array<GKey>{ GKey::SPACE });
 	map.set("shoot", Array<GKey>{ GKey::LEFT_MOUSE });
 	map.set("scope", Array<GKey>{ GKey::RIGHT_MOUSE});
-	map.set("dummyShoot", Array<GKey>{ GKey::LSHIFT });
```
bboudaoud-nv, line 16: We generally try to keep historical FPSci behavior the same unless  a new feature demands a change. Since enabling sprint is optional I would suggest not overriding the default `dummyShoot` as many experiments rely on this. Perhaps sprint could be `TAB` to start with.

source/FpsConfig.h
```diff
@@ -71,7 +71,12 @@ class PlayerConfig {
 	Vector2			turnScale = Vector2(1.0f, 1.0f);			///< Player (horizontal/vertical) turn rate scaler
 	Array<bool>		axisLock = { false, false, false };			///< World-space player motion axis lock
 	bool			stillBetweenTrials = false;					///< Disable player motion between trials?
-	bool			resetPositionPerTrial = false;				///< Reset the player's position on a per trial basis (to scene default)
+	bool			resetPositionPerTrial = false;				///< Reset the player's position on a per trial basis (to scene default)'
+	float			sprintMultiplier = 2.0f;					///< Multiplies with players move rate when player is sprinting (defaults to 2x)
+	float			headBobAmplitude = 0.1f;					///< Determines how high/low players head will oscillate during movement (defaults to 0.1)
+	float			headBobFrequency = 1.0f;					///< Determines how fast players head will oscillate during movement (defaults to 1)
+	float			movementAcceleration = 0.05f;				///< Player acceleration during movement (defaults to 0.05)
+	float			movementDeceleration = 0.15;				///< Player deceleration while stopping (defaults to 0.15)
```
bboudaoud-nv, line 79: One more time same as above, I'd start w/ this disabled. Also make line 79 `0.15f`.

source/PlayerEntity.cpp
```diff
@@ -90,21 +90,46 @@ void PlayerEntity::onPose(Array<shared_ptr<Surface> >& surfaceArray) {
...
+		walkSpeed = *moveRate * *sprintMultiplier * units::meters() / units::seconds();
+	}
 
 	// Get walking speed here (and normalize if necessary)
-	Vector3 linear = Vector3(ui->getX()*moveScale->x, 0, -ui->getY()*moveScale->y);
-	if (linear.magnitude() > 0) {
-		linear = linear.direction() * walkSpeed;
+	m_linearVector = Vector3(ui->getX()*moveScale->x, 0, -ui->getY()*moveScale->y);
+	if (m_linearVector.magnitude() > 0) {
+		m_acceleratedVelocity = lerp(m_acceleratedVelocity, walkSpeed, *movementAcceleration);
```
bboudaoud-nv, line 107: What does this imply the units of acceleration are? We should certainly document acceleration w/ correct units. It might be easier to apply a m/s^2 acceleration here as opposed to doing a lerped value. My concern here is that this makes acceleration/deceleration tied to frame rate (number of times this function is called per second).

source/PlayerEntity.cpp
```diff
@@ -147,6 +171,23 @@ void PlayerEntity::onSimulation(SimTime absoluteTime, SimTime deltaTime) {
...
+	if (!m_gettingMovementInput)
+	{
+		if(movementDeceleration != nullptr)
+		m_acceleratedVelocity = lerp(m_acceleratedVelocity, 0.0f, *movementDeceleration);
+
+		if (m_acceleratedVelocity > 0) {
+			m_linearVector = m_acceleratedVelocity * m_lastDirection;
+		}
+
+		m_headBobCurrentHeight = lerp(m_headBobCurrentHeight, 0.0f, 0.1f);
```
bboudaoud-nv, line 184: Again this fixed lerp will likely be dependent on frame rate (# of times this function is called per second). We should work to make this more uniform by using a time-based metric for interpolation.

bboudaoud-nv (changes requested):
Generally speaking this is a nice solid improvement to FPSci player movement mechanics. It would be interesting to see how something like head bob might impact moving player aim strategies.

I would like to see more comments incorporated into the `PlayerEntity::updateFromInput()` code explaining what various conditions/`lerp()` calls are handling. The primary bug I would like to see fixed is the dependence of various parameters on the call rate of `PlayerEntity::updateFromInput()` rather that the simulation time delta, which should be used instead. Along these lines the positional update would preferably occur in `PlayerEntity::onSimulation()` where you have access to the `sdt` value.

I also think we should align on default values that maintain the historical operation of FPSci so head bob/sprint off (always the preferred approach), there are several ways to do this..

source/FpsConfig.h
```diff
@@ -71,7 +71,12 @@ class PlayerConfig {
 	Vector2			turnScale = Vector2(1.0f, 1.0f);			///< Player (horizontal/vertical) turn rate scaler
 	Array<bool>		axisLock = { false, false, false };			///< World-space player motion axis lock
 	bool			stillBetweenTrials = false;					///< Disable player motion between trials?
-	bool			resetPositionPerTrial = false;				///< Reset the player's position on a per trial basis (to scene default)
+	bool			resetPositionPerTrial = false;				///< Reset the player's position on a per trial basis (to scene default)'
+	float			sprintMultiplier = 2.0f;					///< Multiplies with players move rate when player is sprinting (defaults to 2x)
```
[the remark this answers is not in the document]
jspjutNV: I like having sprint disabled by default too.

source/FpsConfig.h
```diff
@@ -71,7 +71,12 @@ class PlayerConfig {
...
+	float			headBobAmplitude = 0.1f;					///< Determines how high/low players head will oscillate during movement (defaults to 0.1)
+	float			headBobFrequency = 1.0f;					///< Determines how fast players head will oscillate during movement (defaults to 1)
```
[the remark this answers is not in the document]
jspjutNV: If there's a toggle to turn off headbob that indicates that head bob is off by default, I'm okay with sensible defaults for the bob parameters. It's much easier for people to start with something sensible when they turn on head bob than to have to manually adjust the parameters every time they try to turn it on.

Of course if we want it always enabled and to use specific values here to "disable" it, then the values should be disabled, but then we'll want an example in the docs of good default values and maybe even a sample in `data-files/samples`.

source/KeyMapping.cpp
bboudaoud-nv: The argument against changing defaults is this will break compatibility of old experiments. I agree `LSHIFT` is common for sprint, but none of our old experiments rely on this while many rely on `LSHIFT` for `dummyShoot`. I suppose another way to do this would be to default sprint-mode to off and have both `dummyShoot` and `sprint` map to `LSHIFT`.

@joohwankimNV for input.

source/FpsConfig.h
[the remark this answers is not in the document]
bboudaoud-nv: I'm fine with the enabled flag, though the historical model in FPSci is to set parameter values to 0 to disable effects instead of creating an additional boolean toggle (unless required). For example setting `moveRate = 0` to disable motion instead of having an additional `playerMoveEnabled` flag.

source/KeyMapping.cpp
jspjutNV: Yes, my proposal is to overload `LSHIFT` and have sprint disabled by default. So it's mapped to `LSHIFT` it just doesn't do anything in the default setting until enabled.

It shouldn't break any compatibility by doing that since we aren't changing the old defaults, just mapping another key to it that does nothing by default.

source/FpsConfig.h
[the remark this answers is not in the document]
jspjutNV: There's a mix of `enable` and set-to-0 parameters in FPSci. I wouldn't say that one happens more commonly than others and it's more of a "what makes the most sense" decision.

I'd almost argue that `moveRate` is a special case since movement is an "always on" function while head bob seems more like an optional thing that I'd prefer to fully disable (execute no code) when it's off. Though the code may be inconsequential in runtime at the moment, so it probably doesn't matter.

Another example is the HUD where we don't just set the alpha to 0 to disable it, but we have a toggle to disable everything at once, then a bunch of parameters to customize it when on.

jspjutNV:
> 1. You may want to add a `crouchMultiplier` or `crouchSpeed` to allow designers to reduce player speed when crouched
I agree.

> 2. It may also make sense to add a `sprintWhileCrouched` option to disable sprinting when crouched (this would be the "normal" game dynamic, so default to `False`?)
I believe the normal way it works if you press sprint while crouched is that the crouch is cancelled at least and some games switch to walking speed while others switch to running. I personally lean toward "do what I said most recently" and would switch to running speed, but I can see either making sense. The uncommon thing would be to just keep crouching after pressing sprint. Though now that I think of it, the games I'm thinking about are all toggle to crouch. I'm not sure how to break the tie if both crouch and sprint are pressed at the same time with a hold to crouch and hold to sprint function.

> 3. I'd be interested in comparing the idea of sprint multipler to a constant specified `sprintSpeed` in m/s (I'm realizing `moveRate` should really be renamed to `moveSpeed` while I type this, @jspjutNV for input)
I can see reasons why someone might want to set speeds directly, or might want to do multipliers on top of a base speed. Using something like `baseMoveSpeed` with `runSpeedMultiplier`, `sprintSpeedMultiplier`, and `crouchSpeedMultiplier` (maybe even `proneSpeedMultiplier`?) would give you the chance to assign scale factors and globally change all of the speeds with 1 value. Of course someone might dial in a particular speed they want like sprint and then want to dial in the others, but the coupled design would make that more difficult. I don't have a good feeling for which method of design is more likely for experiments, so I'm okay with either. 

I agree with the renaming of `moveRate`, and maybe it should even become `moveSpeedmps` to try to get "m/s" in there, but I'm not sure those units are totally clear in the name. In general, it's nice to have units in the variable name when there are units to make it easier for people to know what they're setting without going back to the docs every time.

source/KeyMapping.cpp
Tokey: I've made the necessary changes and pushed them. Let me know if they look okay or not. I have changed 'dummyShoot' back to LSHIFT and kept sprinting to LSHIFT as well.

source/FpsConfig.h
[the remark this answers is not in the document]
Tokey: Sprint will now be disabled by default.

source/FpsConfig.h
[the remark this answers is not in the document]
Tokey: I have created flags for both Acceleration and Headbob and kept them both disabled by default. Let me know if it needs any changes.

source/FpsConfig.h
Tokey: Updated.

source/PlayerEntity.cpp
Tokey: Now the acceleration is not tied to the frame rate. Lerp will be controlled by Delta-Time.

source/PlayerEntity.cpp
```diff
@@ -90,21 +90,46 @@ void PlayerEntity::onPose(Array<shared_ptr<Surface> >& surfaceArray) {
...
-		linear = linear.direction() * walkSpeed;
+	m_linearVector = Vector3(ui->getX()*moveScale->x, 0, -ui->getY()*moveScale->y);
+	if (m_linearVector.magnitude() > 0) {
+		m_acceleratedVelocity = lerp(m_acceleratedVelocity, walkSpeed, *movementAcceleration);
+		m_linearVector = m_linearVector.direction() * m_acceleratedVelocity;
+		m_lastDirection = m_linearVector.direction();
+
+		if (!m_headBobPolarity)
+		{
+			m_headBobCurrentHeight = lerp(m_headBobCurrentHeight, -*headBobAmplitude * 4.0f, *headBobFrequency * m_acceleratedVelocity / 200);
```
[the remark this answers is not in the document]
Tokey: Added comments and tied the lerp to deltaTime.

source/PlayerEntity.cpp
Tokey: Updated.

source/FpsConfig.h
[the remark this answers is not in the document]
bboudaoud-nv: Only (super minor) remaining fix here is you added a ''' at the end line 74 you can remove

source/PlayerEntity.cpp
```diff
@@ -147,6 +155,58 @@ void PlayerEntity::onSimulation(SimTime absoluteTime, SimTime deltaTime) {
...
+		}
+
+		if (headBobEnabled!= nullptr && *headBobEnabled) {
+			m_headBobCurrentHeight = lerp(m_headBobCurrentHeight, 0.0f, 0.1f);
+		}
+	}
+	else {
+
+		if (*accelerationEnabled) {
+			m_acceleratedVelocity = lerp(m_acceleratedVelocity, m_walkSpeed, *movementAcceleration * deltaTime);
```
bboudaoud-nv, line 179: I don't believe the units on this math work out correct for `movementAcceleration`/`movementDeceleration` being provided in m/s^2. `lerp(a,b,c)` does `a + (b-a)*c` so if `c` was provided in m/s^2 * s = m/s then the `(b-a)*c` term would have units m/s * m/s = (m/s)^2 which is not correct.

Currently I believe this results in a sort of logarithmic acceleration profile. If `deltaTime*movementAcceleration/Deceleration` is always <1 this becomes asymptotic to the specified move speed (the speed will never actually reach the rate). 

If it were me I'd move away from using `lerp()` here and instead use:
`m_acceleratedVelocity = min(m_acceleratedVelocity + movementAcceleration * deltaTime, m_walkSpeed)`
and
`m_acceleratedVelocity = max(m_acceleratedVelocity - movementDeceleration * deltaTime, m_walkSpeed)`
this should guarantee linear acceleration up to the min/max velocity.

Explainability of exact mechanics of systems in FPSci has become a top priority for us. We started w/ a more "looks right" model but have been bitten by it several times when it comes to explaining exactly how a mechanic worked for reproducibility in a paper.

source/PlayerEntity.cpp
```diff
@@ -147,6 +155,58 @@ void PlayerEntity::onSimulation(SimTime absoluteTime, SimTime deltaTime) {
...
+		else {
+			m_acceleratedVelocity = m_walkSpeed;
+		}
+
+		m_linearVector = m_linearVector.direction() * m_acceleratedVelocity;
+		m_lastDirection = m_linearVector.direction();
+
+		if (headBobEnabled != nullptr && *headBobEnabled) {
+			if (!m_headBobPolarity) {
+				m_headBobCurrentHeight = lerp(m_headBobCurrentHeight, -*headBobAmplitude, *headBobFrequency * m_acceleratedVelocity * deltaTime);
```
bboudaoud-nv, line 190: Again, it is not clear to me how this `lerp()` is working. I assume the theory is that it would scale with number of "footsteps per meter" which in turn scales with speed (higher speed ==> fewer steps per meter) but I'd like to either see more comments here explaining this or something introduced in the documentation describing how it relates.

Ideally you'd be able to set the head bob in a fixed unit that relates to speed (say bobs/(m/s)) and have this code be more deterministic, but for head bob I'm okay with a more fluid approach. Still I'm not sure this does what we want, more displacement (faster speed/longer sdt) implies stronger interpolation of the bob here where my intuition would be the opposite... Maybe I am confused.

docs/general_config.md
```diff
@@ -282,6 +286,9 @@ Note: The sound played when `referenceTargetPlayFireSound` is set to `true` is t
 |`playerGravity`     |m/s^2| The graivty vector that impacts the player                                         |
 |`disablePlayerMotionBetweenTrials`|`bool`|Don't allow the player to move when not in a trial?                  |
 |`resetPlayerPositionBetweenTrials`|`bool`|Respawn the player to their original position between trials?        |
+|`headBobEnabled`    |`bool` | Whether or not the headbob effect is enabled.                                    |
+|`headBobAmplitude`  |m      | How high/low the camera Y will go during headBob motion.                         |
+|`headBobFrequency`  |`float`| How fast the headBob will happen. scales up with movement speed.(sinusoidal)     |
```
bboudaoud-nv, line 291: I don't believe your current implementation is sinusoidal, I'd guess it looks like a sort of shark wave from the `lerp()` operations you are performing. Things like this promised in documentation should either be clearly evident in code (i.e. calling `sin()` or tested with unit tests).

docs/general_config.md
```diff
@@ -282,6 +286,9 @@ Note: The sound played when `referenceTargetPlayFireSound` is set to `true` is t
 |`playerGravity`     |m/s^2| The graivty vector that impacts the player                                         |
 |`disablePlayerMotionBetweenTrials`|`bool`|Don't allow the player to move when not in a trial?                  |
 |`resetPlayerPositionBetweenTrials`|`bool`|Respawn the player to their original position between trials?        |
+|`headBobEnabled`    |`bool` | Whether or not the headbob effect is enabled.                                    |
+|`headBobAmplitude`  |m      | How high/low the camera Y will go during headBob motion.                         |
```
bboudaoud-nv, line 290: Would be good to note this is additive to the `crouch/playerHeight` and also that it is ± 1/2 the amplitude not ± full amplitude (what this comment implies to me)

docs/general_config.md
jspjutNV: I assume the head bob changes the reported player position, so you could use the results `.db` files to visualize the shape of the head bob if you want.

The way head bob affects player position is something that we don't need to worry too much about for this PR, but it's probably another fix that'll be wanted before any carefully controlled studies can use it. I'm not sure off-hand if the bob should change the player position (vertically) or if it should just change the camera and keep the player fixed in space. We don't have animated models and maybe never will, but that compounds with everything else. Again, way outside the scope of this PR.

bboudaoud-nv:
@jspjutNV I'm not sure what the `.vs/.idea` files in this branch are, should we remove them or are they something new to VS 2022?

docs/general_config.md
bboudaoud-nv: I don't know if the player position recorded to db includes head bob/crouch effects, would need to look at this, but its a decent suggestion.

I don't require a pure sine-wave for head bob, but we shouldn't document it as sinusoidal if it is not

bboudaoud-nv (changes requested):
This branch is moving in a positive direction, but there are still a number of small fixes, documentation updates, and minor functionality changes/comments that need to be added before it will be merged.

jspjutNV:
> @jspjutNV I'm not sure what the `.vs/.idea` files in this branch are, should we remove them or are they something new to VS 2022?

I agree, these files don't seem needed. @Tokey , could you remove them from the PR? Basically we don't want anything in the `vs/.idea/.idea.FirstPersonScience/.idea/` or similar paths to come with this pull request.

Tokey:
The .idea files came when I opened the project with Rider. I was having issues with VS 2019, and had to re-install G3D and VS 2019 to get it to working again.

source/PlayerEntity.cpp
Tokey: Thank you for the sugeestions. I have made the necessary changes.

source/PlayerEntity.cpp
Tokey: The amplitude is fixed (It will not go higher than what we set). The frequency of the head-bob will change based on players' velocity. So the faster the player moves, the faster his "head/camera" will oscillate. I have updated it with more comments. Let me know if I need to add a few more comments, break it down even further, or change the whole system all-together.

docs/general_config.md
Tokey: I have removed "sinusoidal" from the document. I was under the impression that it might behave like a sin wave, but I probably am wrong.

docs/general_config.md
Tokey: I have updated the documentation for the headbob. Thank you for pointing these out.

source/PlayerEntity.cpp
bboudaoud-nv: I want to point out that the statement "it will not go higher than what we set" isn't necessarily true here. Again `lerp(a,b,c)` works using `a + (b-a)*c` so if `c` is >1 it will produce a value larger than `b` (in this case the `headBobAmplitude`).

Since your `c` (interpolation) term is `headBobFrequency  * acceleratedVelocity * deltaTime` its hard to say this quantity will always be <1. I'd prefer to move to a model here where we can safely guarantee this behavior across all possible values of `headBobFrequency` and `acceleratedVelocity` if possible.

source/FpsConfig.h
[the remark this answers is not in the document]
Tokey: Updated.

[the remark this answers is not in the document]
bboudaoud-nv: It appears the `'` is still at line 74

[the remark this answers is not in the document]
Tokey: Sorry may have pressed undo. Fixed.

PR #399, opened 2022-11-29
Getenv null check

### LINKED ISSUE #398: getenv("USERNAME") failures in some cases
`getenv("USERNAME")` (at line 7 of `SystemInfo.cpp`) sometimes throws access violation exception.
This happened to me when I newly reinstalled my Windows 10.
Seems like a bug on the Windows side, specifically the installation process.

While waiting for a solution, here is how I get around it.
The exception is resolved when I manually set the username by doing `Windows Key + R` -> type `netplwiz` -> double click on my user account -> enter a user name

bboudaoud-nv:
This branch adds support for null checking results of `std::getenv()` calls which can be `null` when the requested values aren't set in the environment variables (as documented [here](https://en.cppreference.com/w/cpp/utility/program/getenv)).

Merging this PR closes #398.

### DIFF
--- a/source/SystemConfig.h
+++ b/source/SystemConfig.h
@@ -14,7 +14,7 @@ public:
 	SystemConfig(const Any& any);
 
 	static SystemConfig load(String filename, bool saveJSON);
-	Any toAny(const bool forceAll = true) const;
+	Any toAny(const bool forceAll = false) const;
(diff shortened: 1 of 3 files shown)


jspjutNV (approved):
Looks good. I assume the switch to `forceAll = false` isn't too important since we don't write out system config Any files often and if we do, we don't care about most of the values.

Issue #39, opened 2019-07-16
Cylinder target


joohwankimNV:
Add a new target shape, a cylinder. This represents the target shape in FPS games more realistically (standing figures).
Note that we want a black boundary as in our current sphere target.

bboudaoud-nv:
We should make this `.Any` configurable by adding a `model` field to the `TargetConfig` similarly to how the `WeaponConfig` model is applied

bboudaoud-nv:
A [recent commit](https://github.com/NVlabs/abstract-fps/commit/ce2b8bac414d75a93e4188ca75d13bd780908db3) adds support for arbitrary target models. I will close this issue for now, as the cylinder model can be supported. Issue #43 still deals with some naming issues regarding use of arbitrary `obj` files.

PR #400, opened 2022-11-30
Add cylinder to targetmodels renamed sample

jspjutNV:
We added a cylinder to the repository in #39 . This PR adds the cylinder to the `targetmodels` sample which is the renamed previous `spheres` sample since it now has another model in it.


bboudaoud-nv (approved):
This looks good. Showing the cylinder and spheres in the same trial makes me realize we didn't quite get the inverse normal shape scaled "right" for the cylinder as its black outline is much thinner than the sphere, but this is an issue for another PR.

PR #401, opened 2022-11-30
Add a two interval forced choice sample

jspjutNV:
This PR adds a sample of 2 interval forced choice. This sample should probably be updated once #392 is merged, but should work fine for earlier versions of FPSci.

### DIFF
--- a/data-files/samples/twoifc.Experiment.Any
+++ b/data-files/samples/twoifc.Experiment.Any
@@ -1,0 +1,211 @@
+// This is a sample of how to implement a 2 interval forced choice experiment.
+// Note that the `Status.Any` is an important part of the design because it controls
+// the order in which the sessions are presented to the user.
+// Also note that the order of stimulus is determinisitic and likely needs additional
+// design to counterbalance and randomize presentation.
+// 
+// There's a variety of targets included even though they're not used to make
+// swapping to other target types easy during testing.
+
+{
+    // global settings
+    showMenuBetweenSessions = False;
+    logToSingleDb = False;
+    clearMissDecalsWithReference = True;
+    showReferenceTargetMissDecals = False;
+
+    scene = {
+        name = "FPSci Simple Hallway";
+        spawnPosition = Point3(46.08, -2.2, 0);
+        spawnHeading = -90;
+    };
+
+    feedbackDuration = 0.0; // Time allocated for providing user feedback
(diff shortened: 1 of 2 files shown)


bboudaoud-nv:
I suggest we wait until merging the `TaskFlow` branch then make this example use tasks to implement the 2IFC model. This sample could still be useful, but I'd like to know more about why certain things are done (i.e., ` logToSingleDb = False` is fairly uncommon, so I'd like to avoid providing samples w/ behavior that is atypical for FPSci configuration).

jspjutNV:
> ` logToSingleDb = False` is fairly uncommon, so I'd like to avoid providing samples w/ behavior that is atypical for FPSci configuration

This came from the `1hit_sa2019` sample that I started this one from. There's probably a bunch of other stuff that could be dropped. I'll clean it up once we get to the task-level version. As is, it's still nice to have this around as an example of the old-style 2 interval forced choice design.

jspjutNV:
76c97aa has a version using the pre-task-level configuration style of specification.

jspjutNV:
As of 2279276 there appears to be an issue updating the status and the results db with this experiment, and I'm not sure what's going wrong. @bboudaoud-nv , the error I'm getting is the following (from `log.txt`):

```
User selected session: 60hz_delay. Updating now...
Warning insert rows with empty row value array ignored!
Warning insert rows with empty row value array ignored!
Warning insert rows with empty row value array ignored!
Warning insert rows with empty row value array ignored!
Creating table Experiments w/ SQL query:CREATE TABLE IF NOT EXISTS Experiments ( description TEXT NOT NULL ,time TEXT NOT NULL ,hash TEXT NOT NULL ,config TEXT NOT NULL );

Error in INSERT INTO statement (INSERT INTO Experiments VALUES('2IFC',): near "s": syntax error
```

The rest of the sql appears to succeed. I didn't go back and check if this was happening before I switched to task level. I suspect this isn't directly related to this experiment, but maybe it's a mistake in the experiment finding an issue we have a work around for.

bboudaoud-nv (approved):
This sample seems fine. I'd be tempted to exclude the `scene` config since it just uses default values, but otherwise seems fine to me.

jspjutNV:
I removed parts of the scene config (left the name for easier adjustment for people who want to change the scene).

I also increased the target size slightly since it was a bit more difficult than the purpose of this sample justified.

PR #402, opened 2022-12-07
Reset weapon cooldowns at the start of kill tests

### LINKED ISSUE #243: `KillTargetFrontHoldclick` test seems to fail about half the time
I've noticed lately that the `KillTargetFrontHoldclick` test has been failing about half of the time.

I'm not sure why this is happening, and I'm not sure this is important to fix yet, but I wanted to open an issue to track the occurence of this problem when running `FPSci.text`. If you learn anything else about this failure, please post here.

### LINKED ISSUE #275: The first, or near first test run sometimes fails

bboudaoud-nv:
This branch resets the weapon cooldown at the start of target kill tests. Since app state isn't reset between tests the weapon cooldown was (in some cases) preventing a fire event from occurring as the weapon is still in its 0.1s cooldown.

Overall we should probably improve test infrastructure by resetting most of this state in the app prior to running each test instead of just relying on the fixture to do this. It is likely a certain sequence of tests could potentially still result in (illegitimate) failures.

Merging this PR closes #275 and closes #243.

### DIFF
--- a/tests/FPSciTests.cpp
+++ b/tests/FPSciTests.cpp
@@ -288,6 +288,7 @@ TEST_F(FPSciTests, KillTargetFront) {
 	EXPECT_EQ(s_app->sess->currentState, PresentationState::trialTask);
 
 	int spawnedTargets = respawnTargets();
+	s_app->weapon->resetCooldown();
 
 	// Kill the front target - just fire
 	zeroCameraRotation();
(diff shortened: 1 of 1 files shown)


jspjutNV (approved):
There's still some random failures on the weapon timing tests, but this PR fixes the big bug with test failures. Thanks for the great fix!

PR #404, opened 2022-12-13
Fixes for reticle preview and `allowReticleChange`

### LINKED ISSUE #361: Interaction between User configured reticle and experiment configure reticle
#350 added the option for the experiment to define a reticle configuration at the experiment or session level to override the user configuration. The experiment can still allow users to modify their reticle, which will lead to a bad user experience. The work around is to not do both of those at the same time, but if it becomes important, we could implement a warning or error message.

bboudaoud-nv:
This branch fixes a buggy interaction between specifying reticle configuration parameters at the experiment/session/trial level and using `allowReticleChange = true` to enable the user menu reticle configuration GUI.

The app now treats specifying any reticle configuration at the experiment/session/trial level as equivalent to setting the corresponding field's `allowReticle[X]Change` parameter to `false`. Thus the only reticle parameters users are free to configure are those not specified as part of experiment configuration. 

Merging this PR closes #361

### DIFF
--- a/docs/general_config.md
+++ b/docs/general_config.md
@@ -474,7 +474,7 @@ In addition to being able to be specified on a per-user basis, the experiment/se
 |`reticleColor`         |`Array<Color4>` |Provides a range of colors over which to set the reticle color as an `Array` w/ 2 elements (min, max)|
 |`reticleChangeTime`    |`float`         |Provides the time (in seconds) for the reticle to change color and/or size following a shot          |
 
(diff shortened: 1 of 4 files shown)


bboudaoud-nv:
A fuller list of changes in this branch include:
- Automatically update the user menu at the beginning of each trial
- Update the reticle preview (to what is actually drawn) on any change to reticle parameters (not just index)
- Implement a `ReticleConfig` `!=` operator to support the item above
- Conditionally show user menu reticle options based on what is specified in the experiment config
- Fix a bug keeping reticle change time from showing when color change was disabled

jspjutNV (approved):
Looks like a good fix and passes all tests. Approved.

PR #405, opened 2022-12-16
Aim assist

bboudaoud-nv:
This branch adds support for a simplified aim assist to FPSci. 

Supported features include:

- An FoV-based aim assist specified by `aimAssistFoV` in general configuration
- Ability to visualize the aim assist FoV using a custom color (`aimAssistFoVColor`) and width (`aimAssistFoVWidth`)
- Automatic invoking of aim assist on fire using `aimAssistSnapOnFire`
- Use of the `autoAim` key (now specified in the key map and defaulting to Tab) to snap aim (when `aimAssistFoV` > 0)
- Optional use of aim assist with reference (`aimAssistOnReference`) and real (`aimAssistOnReal`) targets
- Speed limiting for aim assist using `aimAssistMaxSpeed` parameter (in °/s)


jspjutNV (approved):
I tested this and it seems to work. It would be nice to have a sample to add to the repository, but we can always build that later.

The combo of `aimAssistMaxSpeed` and `aimAssistSnapOnFire` can give some weird behavior, but it's easy for the experiment designer to avoid those situations. We can always improve that behavior later.

PR #406, opened 2023-01-04
Set guard band to 0

jspjutNV:
This PR sets the guard band to 0 instead of 64 like it was previously. This effectively fixes and issue that the configured FoV was 93.75% of the intended value (1920/2048) at 1080p.

I'm not sure if this is the fix we want for this issue yet.

**IMPORTANT: This breaks backwards compatibility with older experiments since the FoV setting gives different results.**

### DIFF
--- a/source/FPSciApp.cpp
+++ b/source/FPSciApp.cpp
@@ -1682,7 +1682,7 @@ FPSciApp::Settings::Settings(const StartupConfig& startupConfig, int argc, const
 
 	useDeveloperTools = startupConfig.developerMode;
 
-	hdrFramebuffer.depthGuardBandThickness = Vector2int16(64, 64);
+	hdrFramebuffer.depthGuardBandThickness = Vector2int16(0, 0);
 	hdrFramebuffer.colorGuardBandThickness = Vector2int16(0, 0);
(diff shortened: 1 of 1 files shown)


bboudaoud-nv (approved):

bboudaoud-nv:
Unless we can think of a place where missing guarbands will create visual artifacts/other issues I'm in favor of merging this ASAP. It fixes discrepancies I was seeing in the `AimAssist` FoV visualization and seems to be a generally positive fix for the incorrect horizontal FoV specification.

jspjutNV:
Okay, I'll do a little testing to see if there are any edge artifacts with the TAA. I know there's other TAA issues, but we might want to avoid any huge additional issues. It's possible it'll be just fine. I'll test now, then merge it if I don't find any problems.

bboudaoud-nv:
As we discussed, I'm not too worried about this introducing issues, it seems like we weren't necessarily filling guardbands in the compositing pipeline anyway, so hopefully this has minimal impacts. I'm happy to wait for more testing before we merge though.

jspjutNV:
I tested and don't really see anything broken. It also passes all tests. I'm merging it.

PR #408, opened 2023-01-05
Fix for crash on load w/ experiment complete

bboudaoud-nv:
After trial-level configuration the `playerCamera` was not initialized in FPSci when loading a completed experiment. This caused null pointer exceptions when running `oneFrame()`. This branch assigns the `playerCamera` and adds null checking to avoid these exceptions while still displaying the `allSessionsCompleteFeedback` message.

### DIFF
--- a/source/FPSciApp.cpp
+++ b/source/FPSciApp.cpp
@@ -643,6 +643,7 @@ void FPSciApp::updateSession(const String& id, const bool forceSceneReload) {
 		// Create an empty session
 		sessConfig = SessionConfig::create();
 		sess = Session::create(this);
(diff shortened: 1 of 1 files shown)


jspjutNV:
I assume this PR is still in draft state just to be sure it's the type of solution to the problem that we really want. If so, I'll start looking at it.

bboudaoud-nv:
Yeah, exactly. I don't think it will create any negative impacts but certainly worth deciding if this is the behavior we want.

jspjutNV (approved):
Looks good and passes all tests. Nice fix!

PR #410, opened 2023-01-09
Session progress fix

bboudaoud-nv:
This branch fixes a bug that appears to have been created when merging trial-level configuration that prevented:

1. Session-level progress from being logged to the `Status.sessions.csv` file
2. Pretrial feedback from being cleared prior to the trial

These issues both seem to be sourced from the changes supporting trial-level config merged in #392.


bboudaoud-nv:
This branch has merged the `2IFC_Sample` for testing, it should be merged back to `master` _after_ an initial merge of that branch or modified to avoid checking in this files from another branch.

bboudaoud-nv:
Most fixes in this branch come from changes in the FPSci state machine management in #392.

For example, the `trialFeedback` state is no longer used to present the reference target/clear feedback messages, thus an explicit clear of feedback needed to be added when leaving the `referenceTarget` state.

Similarly a left-over `moveOn = true` was updating to a new session regardless of whether questions at the end of a session were answered.

Finally an interaction between how the user menu is shown between sessions and a special case for handling showing the menu at startup were interacting to show the user menu between trials in the first session, this has now been corrected.

jspjutNV (approved):
This has fixed the lack of progress issue.

PR #411, opened 2023-01-10
Don't modify widget manager prior to reinit

bboudaoud-nv:
This branch fixes exceptions thrown when changing experiments using the GUI in debug mode. The fix is simply not to modify the `m_widgetManager` prior to reinitializing the application.


jspjutNV (approved):
I'm not sure if this introduces a memory leak, but since it fixes an annoying debug exception, I approve it. Might be worth keeping an eye on if we run into other issues.

PR #414, opened 2023-01-14
Don't load weapon sounds if audio disabled

### LINKED ISSUE #413: Setting `audioEnable` to false in startup config causes a crash
To reproduce this, set `audioEnable` to false in the startup config with any experiment.

bboudaoud-nv:
This branch fixes an issue in the `Weapon::loadSounds()` call that was creating an exception when `audioEnable` was set to `false` in the startup config. 

Merging this branch closes #413.


jspjutNV (approved):
seems good

PR #416, opened 2023-01-18
Experiment selection fixes

bboudaoud-nv:
This branch attempts to fix some remaining bugs that occur when running in the `Debug` configuration with `developerMode = true`. Specifically exceptions were being thrown around removing elements that were not present in the widget manager. This was an issue before and after calling `clear()` on the widget manager was removed.

This branch also resets the `m_firstSession` flag, avoiding issues w/ undesired state when switching experiments. It should be thoroughly tested for behavior changes and possible remaining exceptions before being merged as the case it impacts is relatively niche, but it could break behavior in `Release` if not carefully implemented.


jspjutNV (approved):
These changes look reasonable and it passes all tests.

jspjutNV (changes requested):
Sorry, I was just testing #348 and I'm now getting crashes unrelated to #348 from this branch. It's probably not ready to merge yet.

PR #418, opened 2023-01-18
Check whether a dialog is present before FPM mode

### LINKED ISSUE #417: Interaction of user menu and questions
When opening the user menu during a question/having the menu open prior to a question being asked FPSci can enter a state in which the cursor isn't drawn despite the application asking for an answer to a mouse-based question. Specifically this is reproducible when clicking the `Resume` button in the menu as opposed to closing it with the menu key bind.

You can reproduce this issue by waiting for a question to be asked in any experiment, opening the user menu (with its key bind), then closing the user menu by pressing the `Resume` button. This results in the question still being drawn, but the mouse mode reverting to `MOUSE_FPM`.

We should test exhaustively for ordering here, but perhaps we should move to a unified call to set the mouse mode that checks whether a `dialog` is present as is done in  `FPSciApp::toggleUserSettingsMenu()` before changing the mouse mode.

bboudaoud-nv:
This branch adds a check for whether a dialog is open prior to entering `MOUSE_FPM` mode (first-person view) in FPSci. Currently it only does this for a single path (user menu `Resume` button), but we could extend this to be the way we set the mouse mode more generally in FPSci (i.e. only change the mouse mode to `MOUSE_FPM` when neither a dialog nor the menu is present).

Merging this PR closes #417.

### DIFF
--- a/source/GuiElements.cpp
+++ b/source/GuiElements.cpp
@@ -806,5 +806,6 @@ void UserMenu::setVisible(bool enable) {
 void UserMenu::resumePress() {
 	if (m_config.requireUserAdd && !m_app->userAdded) return;
 	setVisible(!visible());
-	m_app->setMouseInputMode(visible() ? FPSciApp::MouseInputMode::MOUSE_CURSOR : FPSciApp::MouseInputMode::MOUSE_FPM);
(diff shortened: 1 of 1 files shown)


jspjutNV (approved):
Looks reasonable and is unlikely to cause any worse issues than the one it fixes. Approved.

PR #419, opened 2023-01-19
Multishot weapons

bboudaoud-nv:
This branch adds a `pelletsPerShot` parameter to weapons, where all pellets are fired at the same simulation step and each pellet does `damagePerSecond` * `firePeriod` damage to each target it hits. This is to emulate "shotgun-style" weapons which still have a large cooldown but fire multiple pellets in every shot. This change should work for both hit scanned and propagated projectiles.

Due to the nature of the changes in this branch (iterating in the `Weapon::fire()` routine) we should be careful it doesn't have performance impacts for large target/projectile counts and document where it does. The current implementation does run a nested loop that is # of pellets * # of targets total iterations.


bboudaoud-nv:
Before merging this branch we should figure out how we want to log the results of multi-pellet shots

PR #420, opened 2023-02-06
Add adaptive stimulus

bboudaoud-nv:
This branch adds support for adapting FPSci stimulus at runtime using external scripting. This model is further detailed in documentation updates to `experimentConfigReadme.md`. 

A sample should be added to demonstrate the concept before we mark this branch ready for review.

The current approach is not heavily tested for mixing `adaptive` and `constant` tasks in a single session in FPSci as this is not the core use case. Theoretically the approach should work, but there may be additional issues.

We should think about the upsides/downsides of adopting a scripting model like this to run fixed count (`constant`) tasks as well as fully adaptive ones.

### DIFF
--- a/docs/experimentConfigReadme.md
+++ b/docs/experimentConfigReadme.md
@@ -100,11 +100,18 @@ FPSci tasks are a way to affiliate trials into meaningful grouping with repeat l
 
 Tasks are configured using the following parameters:
 - `id` an identifier for the task, used for logging
+- `type` can be `"constant"` for constant stimulus (uses trial orders) or `"adaptive"` for [script-based adaptive stimlulus](#adaptive-stimulus)
+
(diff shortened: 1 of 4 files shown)


bboudaoud-nv:
There are a few topics we should discuss prior to merging this PR, examples include:

1. The possibility of the adaptation script having the results database file open blocking FPSci from writing new results/dropping data.
2. The challenges of relying on/synchronizing results files and system state, particularly for multi-user runs of an experiment on a single machine.
3. The current progress mechanism and if/whether we want to modify it at all to (1) report trial level updates and (2) report overall session-level progress instead of just the current task.
4. Are there any (new) feedback string values we'd like to include in FPSci to help guide subjects through adaptive stimulus trials?

PR #422, opened 2023-03-15
Updates for VS 2022

bboudaoud-nv:
This branch updates the FPSci solution to work with visual studio 2022 (from 2019). In addition to updating the `PlatformToolset` to `v143` this branch also updates the google test package to version `1.8.1.7`.


jspjutNV (approved):
I tested this with G3D r7220 and it appears to work. I also updated the `readme.md` to reflect the new VS and G3D version requirements.

PR #423, opened 2023-03-30
Update on general_config.md

joohwankimNV:
The configuration variable specifying commands seem outdated. I updated it to my knowledge, please double check!


bboudaoud-nv:
This seems to be using the C++ variable names instead of the Any serialized names to get the value from Any. I believe what is documented currently is correct. We can change these names if we think the C++ names are more descriptive.

bboudaoud-nv:
@joohwankimNV can you confirm that this change is unnecessary? If so I'll close this PR.

joohwankimNV:
Yes, I agree this change is unnecessary. We can close this.

Joohwan
________________________________

PR #426, opened 2023-04-12
Fix for `showReferenceTarget = false`

### LINKED ISSUE #425: `showReferenceTarget = false` appears to be broken
It seems that setting `showReferenceTarget` to `false` no longer disables the reference target. This is likely an oversight in FPSci state machine changes.

bboudaoud-nv:
This fix bypasses the `referenceTarget` state of the state machine when `showReferenceTarget = false`, avoiding the issue of ignoring this configuration.

Merging this PR closes #425.


jspjutNV (approved):
This looks reasonable, and passed all tests. Approved.

PR #427, opened 2023-04-14
Added G3D shaders that seem required

jspjutNV:
After updating to the latest G3D r7220, the `log.txt.` identifies a couple more shaders as being used. This PR adds those shaders to the packager script.


bboudaoud-nv (approved):
I haven't looked into what these shaders do, but can confirm they are reported in `log.txt`, so we should include them.

PR #433, opened 2025-01-31
Experiment List Enable startup config option

jspjut:
Adds the `experimentListEnable` startup config option to the startup config. This option is forced on when `developerMode=true` but now can be turned on independently of developer mode if desired.


jspjutNV:
@bboudaoud-nv FYI, I no longer have permissions on this repo to do anything. That means I can't merge this PR, nor can I put in a release with the change.

jspjut:
In some testing, we've found that this does work, but isn't super good when many experiments use the same user config, and you're also trying to add users dynamically since the sessions lists end up being created wrong.

